PR #3, opened 2022-04-28
Enable BINFMT_MISC in the kernel

Lucus16:
Compiling the kernel with BINFMT_MISC allows users to emulate x86
software much like Rosetta on macos with:

    boot.binfmt.emulatedSystems = [ "x86_64-linux" ];


tpwrules:
This looks like a good idea, thanks. Do you know how this interacts with the 4k/16k pages? What is the performance like? Having x86 emulate ARM is pretty miserable, but it is functional.

Lucus16:
A build of GNU hello takes 15 seconds native and 5 minutes when using x86 emulation and 4k pages. It's not fast. I'll try 16k pages later.

PR #6, opened 2022-05-15
Add hint on terminus font

WhyNotHugo:
Thanks for including the package itself on the base image; I would have had no idea of which font to look for and this was a life saver!


tpwrules:
Thank you for the suggestion. I've added some info on how to do this.

PR #7, opened 2022-05-16
change cpufreq governor to schedutil

zzywysm:
On arm64 machines, especially those with E and P cores, it's better to go with "schedutil" not "ondemand".  See https://blog.elementary.io/elementary-os-on-pinebook-pro/ for more context.


nix/m1-support/kernel/config
```diff
@@ -48,7 +48,8 @@ CONFIG_CRASH_DUMP=y
 CONFIG_XEN=y
 # CONFIG_ARM64_PTR_AUTH_KERNEL is not set
 CONFIG_RANDOMIZE_BASE=y
-CONFIG_HIBERNATION=y
+CONFIG_SUSPEND=n
+CONFIG_HIBERNATION=n
```
tpwrules, line 52: I'm unsure how these changes relate to the governor change. Are sleep and hibernation broken in the current kernel? I'd prefer that they be a separate commit if so.

nix/m1-support/kernel/default.nix
```diff
@@ -10,7 +10,7 @@
 
     # set a default frequency governor the same way nixos-generate-config does
     # so the necessary bits get properly cross-compiled
-    powerManagement.cpuFreqGovernor = lib.mkDefault "ondemand";
+    powerManagement.cpuFreqGovernor = lib.mkDefault "schedutil";
```
tpwrules, line 13: Please change the comment to indicate why schedutil was chosen over the default ondemand.

Please also remove the `lib.mkDefault`. If it's present, a user who follows the guide to install a new system will get an error that there is a conflicting definition for the option as `nixos-generate-config` puts ondemand by default.

tpwrules (changes requested):

zzywysm:
I have made a mess of things.  I'll resubmit this in a new pull request.

PR #9, opened 2022-06-21
update kernel energy settings

zzywysm:
We remove hibernation and suspend from the kernel config because the Asahi kernel doesn't support them yet.

We change the cpufreq governor to "schedutil" because it is a prerequisite for using energy-aware scheduling.


tpwrules:
For whatever reason, your proposed kernel config change makes my 16" M1 Max MBP unbootable because it is unable to find the root device whether it is on NVMe or USB. I've taken the frequency governor change by itself but not the kernel change.

PR #10, opened 2022-07-16
kernel: update kernel configuration

zzywysm:
This broadly modifies the Asahi Linux's kernel configuration to remove code not currently needed.  Among many other changes, we remove USB3 and suspend/hibernate because the Asahi Linux team doesn't support them yet.

Tested on my M1 MacBook Air.  Upon boot, this configuration led to 1GB less RAM consumption according to neofetch.


nix/m1-support/kernel/default.nix
```diff
@@ -11,18 +11,18 @@
     # we definitely want to use CONFIG_ENERGY_MODEL, and
     # schedutil is a prerequisite for using it
     # source: https://www.kernel.org/doc/html/latest/scheduler/sched-energy.html
-    powerManagement.cpuFreqGovernor = lib.mkOverride 800 "schedutil";
+    powerManagement.cpuFreqGovernor = lib.mkDefault "schedutil";
```
tpwrules, line 14: This has to stay at `lib.mkOverride 800` to avoid conflicting with the nixos-generate-config default and any user who wants to change this.

tpwrules (changes requested):
Thank you for this. It boots on my machine now. However, you said this configuration targeted a slightly older version of the Asahi kernel. Why don't you wait until I finish the changes for the recent release (which will happen tonight) then rebase on that and add your latest configuration?

I'd eventually like to move in the direction of the stock NixOS config but having a minimal working config is a good thing too.

zzywysm:
> Thank you for this. It boots on my machine now. However, you said this configuration targeted a slightly older version of the Asahi kernel. Why don't you wait until I finish the changes for the recent release (which will happen tonight) then rebase on that and add your latest configuration?

Agreed.  I'll submit another pull request after you move to Linux 5.19-rc6-asahi.

PR #15, opened 2022-08-23
kernel: Fix pure evaluation

zhaofengli:
This fixes usage in pure evaluation mode (i.e., flakes) and doesn't affect any existing use-cases.


tpwrules:
Thank you! This in fact would affect someone who wants to build a cross installer through flakes on a non-aarch64 system but that's a very niche case, certainly more than using their system through flakes. If that person stumbles on this, we can consider changing it.

PR #24, opened 2022-10-16
Add uinput as a module

natsukagami:
I wonder why it was not included, as it was part of https://github.com/AsahiLinux/PKGBUILDs `config`.
I needed it for `input-remapper`.


tpwrules:
Thanks for this. The kernel config is still in a bit of a sorry state and is based on a pretty early and old revision of that `config`.

I would appreciate a better one but I don't know too much about good options. Building a full NixOS kernel with its default config takes forever which is why I haven't done it so far.

luxus:
can you not let ci build it? :D

tpwrules:
That's one of the constraints, CI time is limited to 6 hours on a 2 core processor for the whole image. I admit I haven't tried but based on how long things take on my personal machines a full kernel build could exceed that.

PR #26, opened 2022-11-24
Kernel 6.1 with m1n1 1.1.8

natsukagami:
Work in progress. I'm using the non-edge PKGBUILD right now, except for the `SUSPEND` flag.

Some formatting problem because I'm letting rnix auto-format...

- Keyboard backlight works
- Suspend seems to work
- DCP makes terminals too slow... (`edge` with DCP drivers work though)


natsukagami:
Oh, 4k patch doesn't seem to work anymore.

tpwrules:
Thanks for starting this. Unfortunately I had already started the work yesterday and just had to fix a small issue before pushing today. If you want to fix the 4k patch that would be greatly appreciated. I'm not sure how I want the different kernel configs (plus GPU support) to work yet but input on that would be helpful too.

One small nit: you would have broken the USB installer by dropping `CONFIG_SQUASHFS_ZSTD=y` from the config.

natsukagami:
Thanks for the quick reply! I am not a kernel tinkerer, so I'm afraid there is not much I could do... Just wanted to save you from some manual labor :P 

I'll just switch back to your branch for now. If something comes up later I will open another PR.

PR #27, opened 2022-11-25
Pass `boot.kernelPatches` to the kernel package

natsukagami:
... to make tinkering with the kernel configs and custom patches as simple as following the [wiki](https://nixos.wiki/wiki/Linux_kernel).

I also modified the `readConfig` function a little bit to make it allow config overrides, by basically just build the Nix expression from `//`-ing single-property sets.

Feel free to edit my branch as you need, I might not be around to reply to change requests.

### DIFF
--- a/nix/m1-support/kernel/default.nix
+++ b/nix/m1-support/kernel/default.nix
@@ -4,6 +4,7 @@
 {
   config = {
     boot.kernelPackages = config.hardware.asahi.pkgs.callPackage ./package.nix {
+      inherit (config.boot) kernelPatches;
       _4KBuild = config.hardware.asahi.use4KPages;
     };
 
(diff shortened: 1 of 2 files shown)


tpwrules:
Does this actually work for the `extraConfig` syntax documented on the wiki? It looks like you'd need something more like what goes in the `.config` file. Can you provide a working example?

This would be a nice feature to have though, particularly with the Asahi variants.

natsukagami:
Yikes, I forgot about the differences. I could write a script that translates the `extraConfig` syntax back to the `.config` file syntax.

natsukagami:
I wrote a simple Nix script to parse it, it's not good enough for all the features of the format (I think they allow something like `CONFIG? value` so it's an override-able value, but we just have top-to-bottom overriding). 
I am using it right now to enable suspend on my kernel: https://github.com/natsukagami/nix-home/blob/master/kagami-air-m1/configuration.nix#L26

tpwrules:
I don't love using `//` for thousands of options as I think it might be quadratic complexity, nor hiding the error that a config option has been duplicated at least in the main file. But I recognize this is a useful feature.

PR #31, opened 2022-12-18
software package updates 2022-Dec

zzywysm:
Update to the latest versions of the Asahi downstream Linux kernel, and the m1n1 bootloader.


tpwrules:
Already taken care of and pushed now. Thanks for the effort though.

PR #32, opened 2022-12-18
general improvements 2022-Dec

zzywysm:
Here are a few quality of life improvements for NixOS on Apple Silicon.

- We add in wget, curl, and magic-wormhole into the installer configuration in case a user wants to download a configuration.nix while in the installer mode
- We include the Linux firmware set, primarily because it's easier to plug in a USB-to-Ethernet adapter than type in multiple commands to bring up Wi-Fi
- We ignore the Mach-O build of m1n1 because the Apple bootloader hasn't supported custom Mach-O objects since macOS 12.1
- We enlarge the u-boot font so that text is readable during the boot process


nix/installer-bootstrap/installer-configuration.nix
```diff
@@ -65,11 +65,15 @@
...
+    pkgs.magic-wormhole
   ];
 
-  # save space and compilation time. might revise?
-  hardware.enableAllFirmware = lib.mkForce false;
-  hardware.enableRedistributableFirmware = lib.mkForce false;
+  # add in firmware so usb-ethernet adapters work
+  hardware.enableAllFirmware = lib.mkForce true;
+  hardware.enableRedistributableFirmware = lib.mkForce true;
+  nixpkgs.config.allowUnfree = true;
```
onethirtyfive, line 76: Is allowing unfree software by default (e.g. by maintainers rather than an end-user) controversial or, worse, liability-causing? (Genuine question--I have no idea.)

zzywysm: From the commit message:

"As booting Linux on Apple Silicon already requires many unfree Apple blobs, this change should not offend anyone's sensibilities."

nix/installer-bootstrap/installer-configuration.nix
```diff
@@ -65,11 +65,15 @@
...
     pkgs.cryptsetup
+    pkgs.wget
+    pkgs.curl
+    pkgs.magic-wormhole
   ];
 
-  # save space and compilation time. might revise?
-  hardware.enableAllFirmware = lib.mkForce false;
-  hardware.enableRedistributableFirmware = lib.mkForce false;
+  # add in firmware so usb-ethernet adapters work
```
tpwrules, line 73: How much size does this add to the ISO? IIRC it was quite a lot. But being able to use an Ethernet adapter is useful, maybe we can just pull out that firmware. What chipset is yours?

nix/m1-support/u-boot/default.nix
```diff
@@ -21,20 +21,29 @@ in (buildPkgs.buildUBoot rec {
...
     "u-boot-nodtb.bin.gz"
-    "m1n1-u-boot.macho"
     "m1n1-u-boot.bin"
   ];
   extraConfig = ''
     CONFIG_IDENT_STRING=" ${version}"
+    CONFIG_VIDEO_FONT_4X6=n
+    CONFIG_VIDEO_FONT_8X16=n
+    CONFIG_VIDEO_FONT_SUN12X22=n
+    CONFIG_VIDEO_FONT_TER12X24=n
```
tpwrules, line 31: I do not understand why you add the patch `mps-u-boot-ter12x24.patch` if you do not enable `CONFIG_VIDEO_FONT_TER12X24`?

nix/installer-bootstrap/installer-configuration.nix
```diff
@@ -65,11 +65,15 @@
     pkgs.gptfdisk
     pkgs.parted
     pkgs.cryptsetup
+    pkgs.wget
+    pkgs.curl
+    pkgs.magic-wormhole
   ];
```
tpwrules, line 71: I will have to check how much compilation time these add. In particular magic-wormhole is Python and brings in a bunch of modules. Maybe wormhole-william is a better fit?

nix/m1-support/u-boot/default.nix
```diff
@@ -21,20 +21,29 @@ in (buildPkgs.buildUBoot rec {
...
   '';
 }).overrideAttrs (o: {
   # nixos's downstream patches are not applicable
-  patches = [ ];
+  # however, we add in bigger u-boot fonts because the mac laptop screens are high-res
+  # these patches obtained via:
+  # https://git.alpinelinux.org/aports/tree/testing/u-boot-asahi
+  patches = [ 
+    ./apritzel-first5-video.patch
+    ./mps-u-boot-ter12x24.patch
```
tpwrules, line 41: I would prefer to use `fetchpatch` to download these (ideally from the original source) instead of vendoring them as they are pretty large.

nix/m1-support/m1n1/default.nix
```diff
@@ -70,7 +70,6 @@ in stdenv.mkDerivation rec {
     runHook preInstall
 
     mkdir -p $out/build
-    cp build/m1n1.macho $out/build
```
tpwrules, line 73: Is the `.macho` still useful for the hypervisor and remote download functionality? Doesn't seem harmful to leave it and obviously m1n1 still builds it.

nix/installer-bootstrap/installer-configuration.nix
zzywysm: It adds approximately 500 MB.  Given that the minimum size of a USB flash drive these days is 4 GB, I don't think it's a problem.  (If your concern is the transfer speed, you need to set `bs=1048576` when you invoke `dd`.

The key firmwares to make sure we have are the ones for Realtek RTL815x.  The RTL8152 is inside Apple's gigabit ethernet adapter.

nix/m1-support/u-boot/default.nix
zzywysm: Perhaps someone will prefer that font over the largest one I chose.  I don't think it affects code size, because based on the code I think the font inclusion is guarded by ifdefs based on the config options.

nix/installer-bootstrap/installer-configuration.nix
zzywysm: I trust the crypto behind magic-wormhole and find it an incredibly convenient tool.  I haven't reviewed the code in wormhole-william.

tpwrules: I am a wormhole user too, it's fantastic.

wormhole-william uses the same protocol as and is interoperable with magic-wormhole, so it should be exactly as trustworthy, modulo side-channel attacks or possibly malicious endpoints. But it's written in Go so it's way smaller to install and simpler to compile.

nix/m1-support/m1n1/default.nix
zzywysm: It appears the hypervisor switched over from requiring macho to supporting raw images in https://github.com/AsahiLinux/m1n1/commit/0cf1f393ed1409a0bb80696d3984b03a53db66cf

nix/installer-bootstrap/installer-configuration.nix
zzywysm: I haven't noticed any magic-wormhole build happening.  Either I'm blind, or the derivation is getting downloaded from the hydra cache.

tpwrules: I am certain it will need to be built for the cross-compiled installer. This is what I am most concerned about as there are no ARM github actions runners yet.

nix/m1-support/u-boot/default.nix
zzywysm: I originally wrote this using `fetchpatch`, but I was worried about the alpine testing branch I obtained this code from disappearing suddenly.

nix/installer-bootstrap/installer-configuration.nix
tpwrules: I would say `enableAllFirmware` and `allowUnfree` both need to be off. `enableAllFirmware` is specifically marked as including non-redistributable packages, and I don't want to expose the pre-hosted images to that liability. Additionally, it does not seem to include anything useful to us. This then makes `allowUnfree` unnecessary.

nix/installer-bootstrap/installer-configuration.nix
tpwrules: It looks like it "only" actually adds 400MB, though there is a lot of junk in there we absolutely will not need. I am mostly concerned about build and download time. I will fiddle with this a bit after accepting the PR, it should not be hard to pull out just the relevant bits from `linux-firmware` for common USB LAN adapters.

nix/installer-bootstrap/installer-configuration.nix
tpwrules: `curl` adds no additional overhead because it's already built for something else, but `wget` needs to be compiled additionally so it adds a little bit. It's not a big problem though.

However, `magic-wormhole` does not cross-compile properly so it is a non-starter. `wormhole-william` does though, and like I said it is perfectly compatible with `magic-wormhole`. Please either replace `magic-wormhole` with `wormhole-william` or remove it completely.

nix/m1-support/m1n1/default.nix
tpwrules: Okay sounds good, it's easy to put it back if necessary later.

nix/m1-support/u-boot/default.nix
tpwrules: AFAICT this is not a separate branch but a testing folder. If you pin to a specific commit using the [plain link](https://git.alpinelinux.org/aports/plain/testing/u-boot-asahi/mps-u-boot-ter12x24.patch?id=a211444892fc1fea63151af46d544fda345fa492) I can't see why it should disappear. There are a few places in nixpkgs where alpinelinux's aports repo is `fetchpatch`ed from directly too.

zzywysm:
Due to my weak git skills, I'm closing this pull request and will create a new smaller one against a new branch.

PR #33, opened 2022-12-18
kernel-config-2022-Dec

zzywysm:
- adds two kernel configs, "big" and "small".
- "big" is inspired by the Asahi Linux kernel config, with sanity fixes and additional drivers enabled.  It takes approximately 10 minutes to build.
- "small" has many drivers relating to peripherals disabled.  It takes approximately 5 minutes to build.
- the default config is set to "big"
- note that neither config has accelerated gpu support or Rust enabled yet
- because both configs build most of the Asahi drivers directly into the kernel image instead of as modules, we can simplify the list of available kernel modules in default.nix

Both configs tested on an M1 MacBook Air.

Signed-off-by: Zzy Wysm <zzy@zzywysm.com>


tpwrules:
What is the mechanism to switch kernel configs? Did you test booting them off USB in an installer ISO scenario? I am concerned by your changes to the list of available initrd modules.

zzywysm:
> What is the mechanism to switch kernel configs? Did you test booting them off USB in an installer ISO scenario? I am concerned by your changes to the list of available initrd modules.

The process to switch kernel configs is:
cp asahi-nixos-kernel-config-6.1-big config
or
cp asahi-nixos-kernel-config-6.1-small config

I tested booting the big kernel in the installer ISO scenario.  Booted just fine.

zzywysm:
(It goes without saying that after copying your config choice to `config` you need to do a `sudo nixos-rebuild switch`)

tpwrules:
Oh I'm sorry, I misread what you did. I see now you un-moduled a bunch of things. Does that provide a performance benefit or something? I'm unsure why we wouldn't just use the official Asahi config. I do like the idea of the "small" edition, but keeping it up to date might not be trivial.

zzywysm:
> Oh I'm sorry, I misread what you did. I see now you un-moduled a bunch of things. Does that provide a performance benefit or something? I'm unsure why we wouldn't just use the official Asahi config. I do like the idea of the "small" edition, but keeping it up to date might not be trivial.

A small performance benefit, because each module has to be decompressed before getting loaded into kernelspace.

On my MacBook Air M1, running either of my kernel configs results in a list of loaded modules that's pleasingly less than 10.  (To see your current list of loaded modules, run `sudo lsmod`)

tpwrules:
I have done some pondering and made the executive decision that I don't want to offer configs significantly different from Asahi's official one. Crafting and testing kernel configs is not really in my skillset and I don't want them to become out of date as Asahi moves forward and/or take a lot of my time to update. There have been several issues caused by people fiddling with configs as well, which wastes the time of the Asahi folks.

If there are simple tweaks of a couple of options that would make the NixOS experience specifically better, that's fine. If there are things I can change to make it easier for you to use your own custom config, then please open an issue or file a PR for that instead. (I will try to look into a NixOS option to supply a different file, for example). But I don't think it's practical for me to support wholesale changes for everyone like this.

PR #41, opened 2022-12-23
Improvements 2022-12-2

zzywysm:
Small fixes and changes:
- Stop caring about the m1n1 mach-o binary
- Add curl, wget, and wormhole-william to the installer environment to make it easier to copy files
- Fix the "Error switching console mode" bug
- Increase the size of the u-boot font, which makes the NixOS generation picker readable

Previous review comments at https://github.com/tpwrules/nixos-m1/pull/32


tpwrules:
I rewrote the U-Boot fix to be a fix by default instead of a suggestion for the user's own config. Hope you don't mind.

Thanks for the improvements.

PR #42, opened 2022-12-27
Cosmetics 2022 12

zzywysm:
These are two changes to how the boot process looks on screen.  They are ultimately a matter of taste, but I offer this pull request in case your tastes match mine.

- We enable m1n1's console output to the screen.  It flies by quickly, but I think it looks good.
- We disable u-boot's graphical logo in the upper-right corner of the screen.

### DIFF
--- a/nix/m1-support/m1n1/default.nix
+++ b/nix/m1-support/m1n1/default.nix
@@ -36,6 +36,8 @@ in stdenv.mkDerivation rec {
     fetchSubmodules = true;
   };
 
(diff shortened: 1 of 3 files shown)


tpwrules:
Yeah, these are very personal preference items. I would consider it harder if we didn't have to patch the source to do them, but definitely not now.

Is there a way to make it easier for you to do this in your own config? That patch someone else is working on to rearrange things into a nicer and more overlayable package set would help here a lot I think. Let's see how that turns out.

Issue #45, opened 2023-01-12
KDE + Wayland + experimental GPU driver


bkchr:
Hey did someone get this configuration working? 

Using X11 with the experimental GPU driver works, but wayland fails. It starts, but hangs and I couldn't yet find any indication on what is wrong. I tried to remove almost everything from my config and to only add KDE, but it didn't helped.

tpwrules:
Thanks for discovering, reporting, and providing a solution for this. It is fixed now.

I initially hoped we could mix Mesa versions like this but it's clear now that's not possible.

PR #46, opened 2023-01-13
Override mesa globally to make all packages use the asahi mesa

bkchr:
This is required to make plasma wayland working. Otherwise it would reference the normal mesa libraries and they don't work with asahi gpu driver ;)

Closes: #45


tpwrules:
This seems to be fixable with less overhead using `system.replaceRuntimeDependencies`. I am posting this from a Plasma Wayland session running on the GPU.

bkchr:
Nice solution! Didn't know of this :)

Issue #37, opened 2022-12-19
Improvement 3: Reorganize into two code groups: "modules" and "asahi-overlay" (packages)


onethirtyfive:
**Contingent on #36, which is itself contingent on #34. See 'Improvement 1' for motivations and context for this sequence**

## Reorganize into two code groups: `modules` and `asahi-overlay`

Let's create two directories `modules` and `asahi-overlay` and move everything to its proper location. No changes other than references and file locations are allowed. All commits were tested as specified in #34. This concludes code reorganization for now.

onethirtyfive:
PR is "done" I just have nothing to compare to here since it's contingent on #36 .

Issue #36, opened 2022-12-19
Improvement 2: Distinguish uniformly between packages and modules


onethirtyfive:
**Contingent on #34. Visit it for an overview of what's going on.**

## Next up: Standardize package contents by separating modules from packages

Let's abide by the convention of `default.nix` being an artifact with some specific utility. Configuration--the world of "modules"--is generic and written as modules which _use_ packages. Basically this PR injects a distinction between the two and imposes it  across all our `m1-support` packages uniformly. Each commit was tested as explained in #34.

onethirtyfive:
PR is "done" I just have nothing to compare to here since it's contingent on #34 .

PR #35, opened 2022-12-18
#34 Introduce asahi overlay

onethirtyfive:
#34

PR #47, opened 2023-01-16
Flake support, project reorganization, asahi overlay

### LINKED ISSUE #34: Improvement 1: Standardize package inclusion with an overlay
## Motivations/context (here only, in first issue)

Gotta say it: as more of a NixOS end-user than hacker, I'm grateful for this repo. Thank you for making this possible!

You mentioned edge and GPU support being something you'd have to give some thought to.

I have done a lot of "spiking" to that end in my local repo and ended up with major modifications which _I_ find beneficial, but throwing those at you at once would be presumptuous.

Instead, I invite you on a refactoring process step by step (hence the titled numbering). Each improvement builds on the prior and will have its own PR. If at any point I've lost you or you don't like the direction at least you will have everything to that point. Every commit is "tested" in two ways:

1. `nix-build '<m1-support>' -I m1-support=/home/me/Code/nixos/nixos-m1 -A installer-bootstrap`
2. `nixos-rebuild build -I m1-support=/home/me/Code/nixos/nixos-m1` (where in my `configuration.nix` I have `imports =
    [ ./hardware-configuration.nix <m1-support/nix/m1-support>];`)

I hope this jogs `nixos-m1` _and_ `m1-support` domains adequately. Rebuilding these appears to be a no-op, which I think signifies safety--I'm not substantively changing things.

## Introducing an `asahi-overlay` for both `nixos-m1` and `m1-support`

Let's introduce an `asahi-overlay` which we can use in both `nixos-m1` and `m1-support`. This will remove the repetition and "failure surface area" of invoking sibling packages ad hoc N times. Instead, let's have all `m1-support` packages be ambient members of `config.hardware.asahi.pkgs`. This way, referencing them is as easy as `pkgs.asahi-fwextract`. (Because of laziness, they remain `override`-able! Finally, I got that through my head this time.)

That's all these changes do.

### LINKED ISSUE #45: KDE + Wayland + experimental GPU driver

oati:
Resolves #34 (and #36 #37)
Supersedes #35
Supersedes #46 and closes #45 (probably)

This PR does a couple of things:
- separate the project into 3 directories: `packages`, `nixos-module`, and `iso-configuration`
- define a project-wide asahi overlay in `packages/overlay.nix` and update the nixos module and the installer to use it
  - this adds a `hardware.asahi.overlay` option to the module, which the module automatically appends to the system nixpkgs
- add a `flake.nix`, which replaces `default.nix` and `pins.nix` and outputs
  - the new overlay
  - the packages in the new overlay that were previously exposed by `default.nix`
  - the "m1-support" nixos module
  - an `installer-bootstrap` package for every system supported by nixpkgs, cross-compiling as needed
- fix an issue with the nixos module for flake-based configurations that access `/boot/asahi` as a non-flake input

CI and docs would need to be updated.

- CI changes needed:
  - replace `nix-build -A installer-bootstrap-cross -j2`
    with `nix build .#installer-bootstrap -j2`
    - (may need `--experimental-features 'nix-command flakes'` or configuration changes)
- documentation changes needed:
  - update the `nix-build` commands with flake-based commands
  - make cross-compilation instructions less specific to `x86_64-linux`
  - add basic flake usage to the installation instructions (maybe to the "m1 support updates" section)
    - access the module with `inputs.nixos-m1.nixosModules.default`
    - add `/boot/asahi` as a non-flake input
  - document the asahi overlay

I'm currently using the nixos module output of the flake for my system configuration, but the iso is yet to be tested.


oati:
Hmm, including a mesa override in the overlay causes rebuilds for everything that depends on mesa, regardless of whether the experimental GPU option is actually enabled. The option ends up being kind of silly, since it doesn't actually change anything. I'll fix this tomorrow.

oati:
Actually, I think this is good enough. Overriding `mesa` globally causes so many rebuilds that it probably shouldn't be easy to do so.

The overlay now lets you reference `pkgs.mesa-asahi-edge` from anywhere, so you could just override packages as needed, or add an overlay yourself to override globally:

```nix
nixpkgs.overlays = [ (final: prev: { mesa = final.mesa-asahi-edge; }) ];
```

tpwrules:
While I appreciate the effort, this is a big change for the project and I'm not sure about it.

Primarily, I'm not ready to go flakes-only. Though popular, they are still experimental and require extra effort from users to set up, and the simplicity of the current install is useful. There are certainly things I could adjust for a better flakes experience but I don't regularly use them in this context so some suggestions here would be helpful.

I do think the code needs reorganization and was hoping #35 would provide a good solution to that, but it doesn't seem to have happened. I will probably put together a PR myself for this which cribs ideas from this PR and #35 and solicit feedback from you and the community on if that meets your needs. It's hard to accept a big PR like this one without having fiddled with it myself first.

As for the Mesa fixes, I have a better solution to that I am working on which avoids so much recompilation and hope to get it out soon.

sequencer:
wondering is possible for these tricks to be a part of NixOS/nixos-hardware

oati:
> Primarily, I'm not ready to go flakes-only. Though popular, they are still experimental and require extra effort from users to set up, and the simplicity of the current install is useful. There are certainly things I could adjust for a better flakes experience but I don't regularly use them in this context so some suggestions here would be helpful.

The nixos modle can still easily be used without using flakes.

I decided to go with flakes-only for the iso and package outputs because flakes solve many of the problems that `default.nix` tries to solve in more ad-hoc, partial, and/or impure ways.

Namely,
- flakes use pure evaluation, and thus have much better reproducibility. This helped me find and patch up "leaky impurities" in the project. Flakes enforce
  - pure handling of the lockfile and `system`, ensuring that all impurities are tracked
  - pure `nixpkgs` (preventing `builtins.currentSystem` or `builtins.Env`, etc. from leaking in)
  - outputs are always reproducible, unless you explicitly override the pinned inputs, e.g. `nixos-m1.inputs.nixpkgs.follows "my-nixpkgs";`
- `flake.lock` is automatically managed by the `nix flake` CLI (`nix flake lock`, `nix flake update`)
- flakes follow a standard output schema that integrate with nix-commands, so the output is more standarized/predictable
- flake commands, like `nix build github:tpwrules/nixos-m1#installer-bootstrap`, can be run from anywhere, without the user having to download the repo; so the iso build is a one-liner.

oati:
If you'd prefer, (and if you end up agreeing that flakes would be the best solution,) I could get rid of the `flake-parts` dependency to make the simplest "basic nix code" flake, and try and figure out `flake-compat` to maintain the current installation instructions.

oati:
As for just ensuring a better experience for flake users, the nixos module is easy enough to use as a non-flake input (specifically since the nixos-module does not have any required project depnendencies).

```nix
imports = [
  nixos-m1-non-flake} + "/nixos-module"
];
```

You'd just have to get rid of the `/. +` in https://github.com/tpwrules/nixos-m1/blob/main/nix/m1-support/peripheral-firmware/default.nix#L24 so that `/boot/asahi` doesn't break as a non-flake input.

One small advantage as a nixos module user, however, is be that having the project be a flake would give flake users the ability to easily pin the asahi packages to the "working versions" specified by the nixpkgs in the flake lockfile.

oati:
As for just ensuring a better experience for flake users, the nixos module is easy enough to use as a non-flake input (specifically since the nixos-module does not have any required project depnendencies).

```nix
imports = [
  nixos-m1-non-flake} + "/nixos-module"
];
```

You'd just have to get rid of the `/. +` in https://github.com/tpwrules/nixos-m1/blob/main/nix/m1-support/peripheral-firmware/default.nix#L24 so that `/boot/asahi` doesn't break as a non-flake input.

One small advantage as a nixos module user, however, is be that having the project be a flake would give flake users the ability to easily pin the asahi packages to the "working versions" specified by the nixpkgs in the flake lockfile.

oati:
(to clarify) The flake in this PR *does not require* users to have a flake-based system configuration. The nixos module can still be imported manually, through a channel, through a tarball. The PR merely adds a standardized and convenient way to import the module from a flake.

This PR *does* require either an extra CLI flag `--experimental-features 'nix-command flake'` (or a nix user config enabling these) in order to build packages or the iso with the pinned nixpkgs and rust-overlay.

However, this can be completely worked around with via `flake-compat`, which means users will never have to touch nix flakes in any use case.

tpwrules:
Thanks for the clarification on your intentions and the outcomes. Yeah, I agree the current situation is rather ad-hoc and clunky. You raise some good points about how flakes would make it cleaner. I'm not aware that there are any actual problems here at the moment in particular with regard to purity, but making it impossible to make those kinds of mistakes is good. I don't think `flake-compat` is necessary at this point, I'm not a never-flakes person either.

It's very important to me that a user be able to download an ISO off GitHub, boot it, and install quickly without having to download and rebuild a whole kernel on their system. Nix does make it easy, but it's still not a great time on the lower-end machines. Some of the impurities as I recall were there to make this work.

Please give me a week or so to review and play with this in depth. Please also rebase on top of the latest release. I might do another release in a couple days once the recent nixpkgs staging merge makes it to the NixOS release channels.

tpwrules:
> wondering is possible for these tricks to be a part of NixOS/nixos-hardware

@sequencer Some people have tried this (and moving support for these machines into nixpkgs too) but I haven't kept much track and I don't think it's feasible until the kernel upstreaming matures a little more.

oati:
Ok, I'm running into new issues with the recent changes on `main`: https://github.com/NixOS/nixpkgs/issues/199162 https://github.com/NixOS/nix/issues/5868

`system.replaceRuntimeDependencies` just fails in pure evaluation due to `builtins.storePath`, and therefore on flake-based nixos configurations. I haven't found a workaround.

tpwrules:
Hm, that was another memory around impurity that is now surfacing. I used to have `system.replaceRuntimeDependencies` in the default config for another package. I don't think there is a workaround either.

I can add an option to overlay the new mesa I suppose but that's rather unfair for flakes users because essentially everything needed to be rebuilt.

shaunsingh:
Instead of using `fetchFromGitHub` to grab the source for m1n1/kernel/etc, could we just put those as pinned inputs in the `flake.nix`, that way if users wanted to override the source for e.g. mesa to a newer version or fork for debugging they easily could

oati:
I did consider that option while working on this.

One issue is that it would become much more difficult for non-flake users to incorporate the nixos module into their system.

Flake users can already pretty easily override sources with flake inputs via overlays:

```nix
final: prev: {
  mesa-asahi-edge = prev.mesa-asahi-edge.overrideAttrs (old: {
    src = inputs.mesa-asahi;
  });
}
```

oati:
TODO: Try to get `m1n1` and `u-boot` to cross-compile without using `pkgsCross.aarch64-multiplatform`. Cross-compilation should be handled using the `crossSystem` and `localSystem` arguments of nixpkgs, which we can control from our flake.

tpwrules:
The catch with m1n1 is that it has to be a native build so that the tools like the proxy client and command line come out for x86_64-linux but also a cross build so that the m1n1 stub is built for aarch64-linux. Might be possible to split this into two separate derivations?

I think your solution is workable for U-Boot though, it always expects to be run using an aarch64 stdenv.

The bit with the kernel package is necessary to support the "install a kernel from a cross-compiled ISO without rebuilding it" scenario. The comment might be a little bit incorrect but we do in fact build the config derivation "for" x86_64-linux when cross-building the ISO, even though the contents never change. There might be a better way to do that. The current clause there should be compatible with flakes as it specifically checks for currentSystem to exist and assumes you are running natively if not.

oati:
In the absolute worst case, we could explicitly re-import nixpkgs with `crossSystem` and `localSystem` to avoid problems with `pkgsCross.aarch64-multiplatform` (which is I think itself a re-import of nixpkgs with `crossSystem.config = aarch64-unknown-linux-gnu`, which causes problems because it's not the same as importing with `crossSystem.system = `aarch64-linux`).

As for the kernel config IFD, we can just replace `builtins.currentSystem` with `localSystem` as a pure alternative.

note: nixpkgs uses native builds when `crossSystem` = `localSystem` (in some sense). It doesn't (can't) cross-build from the same platform. (although I think there's an open issue to make it always cross-build?)

tpwrules:
Okay, this should be the final rebase. I am still interested in this, just had to make sure everybody's builds didn't break.

oati:
Some notes about [the default firmware directory](https://github.com/oati/nixos-m1/blob/flake/nixos-module/peripheral-firmware/default.nix#L48-L54):

`builtins.pathExists` is impure, and it (I think) always outputs false in pure mode. https://github.com/NixOS/nix/issues/3998 (stale bot....)

Maybe we can set the default to `"/boot/asahi"` and override this in the installer configuration?

tpwrules:
Yes, the fact that `builtins.pathExists` is always false in pure mode is fine. The documentation covers this case and provides the alternate path for the user to point somewhere their flake can access.

The reason for the two paths is evaluation in the installer ISO. The actual installer gets the firmware by running `asahi-fwextract` [automatically after booting](https://github.com/tpwrules/nixos-m1/blob/fe4fff71fd1dbd8733512d0a2b83a84e3cfdc2bd/nix/installer-bootstrap/installer-configuration.nix#L47), the configuration options don't matter for it. But when the user runs `nixos-install`, their configuration will be evaluated when the ESP is mounted at `/mnt/boot/asahi` (assuming they followed the installation instructions) and that's when the second path option comes into play.

oati:
I've incorporated the recent changes from main. a1e2ef9 might need testing for cross-builds, but I think it should be correct.

flake.lock
```diff
@@ -0,0 +1,115 @@
...
+      },
+      "original": {
+        "dir": "lib",
+        "owner": "NixOS",
+        "ref": "nixos-unstable",
+        "repo": "nixpkgs",
+        "type": "github"
+      }
+    },
+    "nixpkgs_2": {
```
tpwrules, line 70: Unsure what the extra copy of nixpkgs is for? I only want to depend on one. Please make sure sub-flakes follow the main one.

flake.nix
```diff
@@ -0,0 +1,69 @@
+{
+  description = "Apple M1 support for NixOS";
+
+  inputs = {
+    nixpkgs.url = "github:nixos/nixpkgs/nixpkgs-unstable";
```
tpwrules, line 5: I would like to keep the same nixpkgs version that was in `pins.nix` and preserve the comment about the Hydra `mobile-nixos` jobset. This means users/CI won't have to wait forever to cross-compile the basic system.

It would break `nix flake update` though since that would be a no-op. Maybe it is better to document how to upgrade to a specific commit (I am not sure myself)?

flake.nix
```diff
@@ -0,0 +1,69 @@
...
+          };
+
+          nixosModules = rec {
+            m1-support = ./nixos-module;
+            default = m1-support;
+          };
+        };
+
+        # build platforms supported for uboot in nixpkgs
+        systems = [ "aarch64-linux" "x86_64-linux" "i686-linux" ];
```
tpwrules, line 26: We can drop `i686-linux` here. The comment is useful though, if it does not evaluate properly on other systems.

iso-configuration/default.nix
```diff
@@ -3,22 +3,23 @@
...
+    "./m1-support/nixos-module"
   ];
 
   # copy the m1-support and installer configs into the iso
   boot.postBootCommands = lib.optionalString config.installer.cloneConfig ''
     if ! [ -e /etc/nixos/m1-support ]; then
+      mkdir -p /etc/nixos/m1-support
       cp ${./installer-configuration.nix} /etc/nixos/installer-configuration.nix
-      cp -r ${../m1-support} /etc/nixos/m1-support
+      cp -r ${../packages} ${../nixos-module} -t /etc/nixos/m1-support
```
tpwrules, line 22: This needs to copy each module into the `m1-support` folder by name directly. Currently the store hashes get included in the path name so the user will end up with a bunch of copies of the two folders instead of overwriting the old one.

We also need a `default.nix` which simply includes the module directory so that the user can just add the `m1-support` directory to their config.

Alternatively I would prefer to migrate both the `nixos-module` and `packages` folders into an `m1-support` folder in the repo (and add that `default.nix` so that channel-based upgrades and people with it checked out would continue to work.

nixos-module/default.nix
```diff
@@ -7,14 +7,22 @@
...
+      hardware.asahi.pkgs =
+        if cfg.pkgsSystem != "aarch64-linux"
+        then
+          import (pkgs.path) {
+            crossSystem.system = "aarch64-linux";
+            localSystem.system = cfg.pkgsSystem;
+            overlays = [ cfg.overlay ];
+          }
+        else pkgs;
+    };
```
tpwrules, line 25: Unsure if the `pkgsSystem` option below is actually working properly. Seemed to want to cross build the whole system instead of just the kernel in the installer, but it's possible I installed it wrong. I need to re-check this.

flake.nix
```diff
@@ -0,0 +1,69 @@
+{
+  description = "Apple M1 support for NixOS";
+
+  inputs = {
+    nixpkgs.url = "github:nixos/nixpkgs/nixpkgs-unstable";
+    rust-overlay.url = "github:oxalica/rust-overlay";
+    flake-parts.url = "github:hercules-ci/flake-parts";
```
tpwrules, line 7: I would prefer a vanilla flake. Do we need the extra abstraction and dependency?

tpwrules (changes requested):
Thanks again for this work. I've been able to build and boot the cross ISO, and have built (but not tried to boot) the ISO natively. Actually installing doesn't quite work correctly yet but I am assuming the packages all still function.

We still need to address documentation too.

flake.lock
oati: The extra copy is from `rust-overlay`. It's not actually used in `overlays.default`, so if you want `flake.lock` to be absolutely minimal, I can make `rust-overlay` a non-flake input.

flake.nix
oati: You can pin a specific commit with
```nix
    nixpkgs.url = "github:nixos/nixpkgs/nixpkgs-unstable/[commit]";
```
or (I think)
```nix
    nixpkgs = {
      url = "github:nixos/nixpkgs/nixpkgs-unstable";
      rev = "[commit]";
    };
```

flake.nix
oati: This can make (especially) larger projects cleaner and easier to manage, but since our `flake.nix` is simple, a vanilla flake should be good enough.

iso-configuration/default.nix
oati: > Currently the store hashes get included in the path name so the user will end up with a bunch of copies of the two folders instead of overwriting the old one.

What do you mean by this?

My intent with the installer command was to preserve the project structure so that relative paths (like `../packages/overlay.nix`) would work, and so that channel or tarball setups would need minimal changes.

The new "module usage instructions" would be to import `m1-support/nixos-module` instead of `m1-support`.

If you just want to make sure that existing setups won't break, we could make a trivial `m1-support/default.nix`, maybe with a deprecation warning. (Is this what you meant in the second paragraph?)

My reasoning in keeping `nixos-module` (formerly `m1-support`) and `packages` separate was that `packages` is sort of independent from our nixos-module. It defines our overlay, which is used for both the nixos module and for flake outputs (m1n1 and u-boot), and it makes sense outside of the context of nixos configurations.

nixos-module/default.nix
oati: Does it do this if you set `pkgsSystem` to `aarch64-linux`?

flake.lock
tpwrules: No, that's not necessary. You should be able to use only one version with something like `inputs.rust-overlay.inputs.nixpkgs.follows = "nixpkgs";`

iso-configuration/default.nix
tpwrules: The problem is that right now, in the installer, there are folders `/etc/nixos/m1-support/<random hash>-packages` and `/etc/nixos/m1-support/<random hash>-nixos-module` so the relative paths are broken. You have to specify the destination explicitly in `cp` to prevent this and copy the two directories in two separate commands.

I agree with the division into two separate folders. But right now it is `<repo>/iso-configuration`, `<repo>/nixos-module` and `<repo>/packages`. I would like it to be `<repo>/iso-configuration`, `<repo>/m1-support/nixos-module` and `<repo>/m1-support/packages`. That way only one folder has to be copied, both in the installer above and by the user to upgrade. That folder contains everything necessary for the module to work.

Additionally, I would like `<repo>/m1-support/default.nix` to be a simple module which just imports `<repo>/m1-support/nixos-module`. I don't think a deprecation warning is necessary, I can't think of a reason to remove this file. This is also necessary for a user who copies the new project structure over the old one to get correct operation and use the new folders.

flake.lock
oati: Looking into it more, I think `rust-overlay` as a non-flake input makes more sense. The `overlays.default` flake output is just `import ./.`, and the overlay itself doesn't depend on nixpkgs or flake-utils.

iso-configuration/default.nix
oati: (Another thing to consider is that we might want to add "M2 support" pretty soon, and we'd want a good directory structure in place for that.)

oati: Another thing to consider is that this repo is for Apple silicon support in general, and I think we should be moving away from referencing "M1," especially with M2 already released.

oati: Maybe we could use `nixos-module/modules` and `nixos-module/packages`?

tpwrules: What about `as-support/modules` and `as-support/packages`? `nixos-as` is a bit generic though but it's the best I have. I don't think it would be appropriate to call it `nixos-asahi`.

oati: Maybe `apple-silicon-support` or `nixos-apple-silicon`?

Why wouldn't `nixos-asahi` work? [Asahi Linux's about page](https://asahilinux.org/about/) describes itself as "an overall project to develop Linux support for Apple Silicon Macs," and it happens to support an Arch distro under the same name.

If all of Asahi gets upstreamed, then Asahi will no longer exist, and this repo would not need to exist.

tpwrules: I like `apple-silicon-support`, with subdirectories `modules` and `packages`, and the `default.nix`.

Regarding Asahi, while I use their work and credit them (and even contribute a little), I don't want to use their name for my project. I don't want to misrepresent that this is officially created or sanctioned or supported by them.

oati: I've pushed a commit with `apple-silicon-support`.

flake.nix
```diff
@@ -0,0 +1,73 @@
+{
+  description = "Apple M1 support for NixOS";
```
tpwrules, line 2:
```suggestion
  description = "Apple Silicon support for NixOS";
```

flake.nix
```diff
@@ -0,0 +1,73 @@
...
+    let
+      # build platforms supported for uboot in nixpkgs
+      systems = [ "aarch64-linux" "x86_64-linux" ]; # "i686-linux" omitted
+
+      forAllSystems = inputs.nixpkgs.lib.genAttrs systems;
+    in
+      {
+        overlays = rec {
+          asahi-overlay = import ./apple-silicon-support/packages/overlay.nix;
+          default = asahi-overlay;
```
tpwrules, line 27:
```suggestion
          apple-silicon-overlay = import ./apple-silicon-support/packages/overlay.nix;
          default = apple-silicon-overlay;
```

flake.nix
```diff
@@ -0,0 +1,73 @@
...
+        };
+
+        nixosModules = rec {
+          m1-support = ./apple-silicon-support;
+          default = m1-support;
```
tpwrules, line 32:
```suggestion
          apple-silicon-support = ./apple-silicon-support;
          default = apple-silicon-support;
```

iso-configuration/default.nix
```diff
@@ -3,22 +3,23 @@
...
   ];
 
   # copy the m1-support and installer configs into the iso
   boot.postBootCommands = lib.optionalString config.installer.cloneConfig ''
     if ! [ -e /etc/nixos/m1-support ]; then
+      mkdir -p /etc/nixos/m1-support
       cp ${./installer-configuration.nix} /etc/nixos/installer-configuration.nix
-      cp -r ${../m1-support} /etc/nixos/m1-support
+      cp -rT ${../apple-silicon-support} /etc/nixos/apple-silicon-support
     fi
```
tpwrules, line 23: This section should be purely a rename, we don't want to add the `mkdir` or change the `cp` flags now that we are back to a single directory.

tpwrules (changes requested):
The rename is not complete. If you want I can push to your branch, or you can fix it. Then I will test again.

apple-silicon-support/modules/peripheral-firmware/default.nix
```diff
@@ -43,20 +44,15 @@
...
           # path when the system is mounted in the installer
           "/mnt/boot/asahi"
         ];
 
-        validPaths = (builtins.filter
-          (p: builtins.pathExists (p + "/all_firmware.tar.gz"))
-          paths) ++ [ null ];
-
-        firstPath = builtins.elemAt validPaths 0;
-      in if firstPath != null then "${/. + firstPath}" else null;
```
tpwrules, line 59: This has needs to be something like
```
default = let
  firmwarePath = lib.findFirst (path: builtins.pathExists (path + "/all_firmware.tar.gz")) null
        [
          # path when the system is operating normally
          "/boot/asahi"
          # path when the system is mounted in the installer
          "/mnt/boot/asahi"
        ];
  in if firmwarePath != null then "${/. + firmwarePath}" else null;
```
so that the firmware is copied into the store and so that we can cope with a missing firmware.

apple-silicon-support/packages/linux-asahi/default.nix
```diff
@@ -15,7 +18,7 @@
        "CONFIG_${builtins.elemAt kv 0}=${builtins.elemAt kv 1}";
     in lib.strings.concatMapStringsSep "\n" perLine lines;
 
-  readConfig = configfile: import (localPkgs.runCommand "config.nix" { } ''
+  readConfig = configfile: import (runCommand "config.nix" { } ''
```
tpwrules, line 21: This has definitely broken `hardware.asahi.buildPkgs`. I am pretty sure the above comment is wrong. The real issue is that the `config.nix` file does not make its way to the ISO. Thus, when the configuration is evaluated, the `runCommand` needs to be executed to get the complete source code, but `runCommand` expects to run on a `hardware.asahi.buildPkgs` system and fails.

I want to keep the `localPkgs` hack for now (we don't need `lib = localPkgs.lib` though) and fix it later. That option will still be broken as it was for flakes users before, but it will be fine for the install scenario.

tpwrules (changes requested):
Please address the mentioned issues. I will then squash, merge, and update the documentation.

nixos-module/default.nix
tpwrules: That works fine. See the latest review.

apple-silicon-support/modules/peripheral-firmware/default.nix
oati: I think just removing the quotes will have the intended effect:
```nix
      default = lib.findFirst (path: builtins.pathExists (path + "/all_firmware.tar.gz")) null
        [
          # path when the system is operating normally
          /boot/asahi
          # path when the system is mounted in the installer
          /mnt/boot/asahi
        ];
```
(since the purpose of `/. +` seems to just be to convert from a string to a path)

tpwrules: This solution seems to work too. Thanks for the suggestion.

ingenieroariel:
Thanks for all this work, I have replaced my local config with the latest version of this PR, at the end of this comment I paste a copy of my flake.nix on /etc/nixos.

FYI, The overlay strategy with the latest version of this flake gives:
```
error: infinite recursion encountered

       at /nix/store/mhx0418frlp5p3i20bk6b0m41klps12m-source/apple-silicon-support/packages/mesa-asahi-edge/default.nix:5:2:

            4|
            5| (mesa.override {
             |  ^
            6|   galliumDrivers = [ "swrast" "asahi" ];
(use '--show-trace' to show detailed location information)
```


```
{
  description = "A very basic flake";

  inputs = {
    nixpkgs.url = "github:nixos/nixpkgs/master";
    home-manager = {
      url = github:nix-community/home-manager;
      inputs.nixpkgs.follows = "nixpkgs";
    };
    rust-overlay = {
      url = github:oxalica/rust-overlay;
      inputs.nixpkgs.follows = "nixpkgs";
    };

    nixos-m1-overlay = {
      url = github:oati/nixos-m1/flake;
      inputs.nixpkgs.follows = "nixpkgs";
      inputs.rust-overlay.follows = "rust-overlay";
    };

    };

  outputs = {
    self,
    nixpkgs,
    rust-overlay,
    nixos-m1-overlay,
    home-manager,
  }: let
    system = "aarch64-linux";
    
    pkgs = import nixpkgs {
      inherit system;
      config.allowUnfree = true;
      overlays = [ 
          rust-overlay.overlays.default
          nixos-m1-overlay.overlays.default
      ];
    };
    lib = nixpkgs.lib;
  in {
    nixosConfigurations = {
      nixos = lib.nixosSystem {
        inherit system;
        modules = [
          ./configuration.nix
          nixos-m1-overlay.nixosModules.default
          home-manager.nixosModules.home-manager
          {
            home-manager.useGlobalPkgs = true;
            home-manager.useUserPackages = true;
          }
        ];
         };
    };
  };
}
```

apple-silicon-support/packages/overlay.nix
```diff
@@ -0,0 +1,8 @@
+final: prev: {
+  linux-asahi = final.callPackage ./linux-asahi { };
+  m1n1 = final.callPackage ./m1n1 { };
+  uboot-asahi = final.callPackage ./uboot-asahi { };
+  asahi-fwextract = final.callPackage ./asahi-fwextract { };
+  mesa-asahi-edge = final.callPackage ./mesa-asahi-edge { };
```
tpwrules, line 6: This needs to be `mesa-asahi-edge = final.callPackage ./mesa-asahi-edge { inherit (prev) mesa };` to prevent infinite recursion if the overlay installation method is used.

oati: I think setting `mesa = final.mesa-asahi-edge.override { inherit (prev) mesa; };` in the mesa module would be better.

oati: @ingenieroariel let me know if the latest commit has any issues

apple-silicon-support/packages/linux-asahi/default.nix
oati: I think the solution is to replace the `"config.nix"` IFD with a pure nix regex parser. It seems like it'll be pretty straightforward, and it can be upstreamed to nixpkgs since nixpkgs also uses an IFD currently.

flake.nix
```diff
@@ -0,0 +1,73 @@
...
+            pkgs = import inputs.nixpkgs {
+              crossSystem.system = "aarch64-linux";
+              localSystem.system = system;
+              overlays = [
+                (import inputs.rust-overlay)
+                self.overlays.default
+              ];
+            };
+          in {
+            inherit (pkgs) m1n1 uboot-asahi;
```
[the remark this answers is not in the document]
oati: The issue I had with outputting the kernel package was https://github.com/NixOS/nix/issues/4265, which also breaks `nix flake show`. Luckily we should be able to get rid of the IFD using pure nix, as I've mentioned below.

We could also output the installer config, but I think it's not as clear where the config should go in the output, especially since we need to account for different possible build platforms, so I wanted to avoid it if possible.

apple-silicon-support/packages/linux-asahi/default.nix
oati: opened #51

oati: I've added back the `localPkgs` workaround for now.

apple-silicon-support/packages/overlay.nix
ingenieroariel: Tried the latest commit and it still complains at:
```

       at /nix/store/pgyn69rm8wv2ppdpz04xxnr68003kphy-source/apple-silicon-support/packages/mesa-asahi-edge/default.nix:1:1:

            1| { lib
             | ^
            2| , fetchFromGitLab

       error: infinite recursion encountered

       at /nix/store/pgyn69rm8wv2ppdpz04xxnr68003kphy-source/apple-silicon-support/packages/mesa-asahi-edge/default.nix:5:2:

            4|
            5| (mesa.override {
             |  ^
             ```

Comment (oati) [reply apple-silicon-support/packages/overlay.nix:6]: I think I'll go with depending on `prev.mesa` then.

(To be honest, I don't think overlays should be a nixos module option in general. It's too dependent on ordering for merging to make sense a lot of the time. I was mostly reluctant to use `prev` because of how little control users will have over the order in which overlays are applied. In my own configs, I just import nixpkgs from my flake, and `mkForce` overlays to `[]`.)

Comment (ingenieroariel) [reply apple-silicon-support/packages/overlay.nix]: As a user, that option would work for me. I guess the main thing is to make overlays possible and documented. Could you share an example flake.nix that does this?

The reason I need it is because firefox is taking 30 seconds on boot, and crashes all the time. I want to rule drivers issues out by compiling it from scratch.

Furthermore, once we get a flake.lock in the repo, we could share artifacts on cachix or a private server to use less electricity.

Comment (oati) [reply apple-silicon-support/packages/overlay.nix]: The latest commit should work now (theoretically), and I'd appreciate it if you could confirm.

<details>
<summary>Here's a simplified version of my NixOS flake.nix. (click to expand)</summary>
<br>

```nix
{
  inputs = { stuff }; # nixpkgs, nixos-m1, etc

  outputs = inputs:
    let
      system = "aarch64-linux";

      pkgs = import inputs.nixpkgs {
        localSystem = { inherit system; };

        config = {
          allowUnfree = true;
        };

        overlays = [
          (final: prev: { }) # bring flake package outputs into pkgs

          inputs.nixos-m1.overlays.default
          inputs.another-flake.overlays.default

          (import ./packages/overlay.nix) # my own package overrides
        ];
      };
    in {
      nixosConfigurations.my-laptop = flakes.nixpkgs.lib.nixosSystem {
        inherit system pkgs;

        modules = [
          # disable module overlays
          { nixpkgs.overlays = inputs.nixpkgs.lib.mkForce []; }

          inputs.nixos-m1.nixosModules.default
          inputs.another-flake.nixosModules.default

          ./configuration.nix
        ];
      };
    };
}
```
</details>
<br>

> Furthermore, once we get a flake.lock in the repo, we could share artifacts on cachix or a private server to use less electricity.

We could use cachix to maybe eliminate the need for `hardware.asahi.pkgs`. I'm not sure how well that would work with non-flake systems, though.

ingenieroariel: I confirm that the change to prev makes the overlay option work.

My computer crashes when compiling firefox but it is unrelated to this, sway and helix worked fine.

ingenieroariel:
I have verified the latest version of this PR on a new (flake-based) configuration and it works.

This set up is very flexible for downstream users like me. Here is an example:

I let nix flakes update the rust overlay to latest and nixpkgs to latest, and it worked after adding this to my configuration.nix (blacklist-types is now blocklist-types).

```
  boot.kernelPatches = [
      {
        name = "edge-config";
        patch = null;
        # derived from
        # https://github.com/AsahiLinux/PKGBUILDs/blob/stable/linux-asahi/config.edge
        extraConfig = ''
          DRM_SIMPLEDRM_BACKLIGHT n
          BACKLIGHT_GPIO n
          DRM_APPLE m
          APPLE_SMC m
          APPLE_SMC_RTKIT m
          APPLE_RTKIT m
          APPLE_MAILBOX m
          GPIO_MACSMC m
          DRM_VGEM n
          DRM_SCHED y
          DRM_GEM_SHMEM_HELPER y
          DRM_ASAHI m
          SUSPEND y
        '';
      }
      {
        name = "blocklist-type";
        patch = builtins.fetchurl "https://patch-diff.githubusercontent.com/raw/AsahiLinux/linux/pull/109.diff";
      }
    ];
```

PR #57, opened 2023-02-17
Fix kernel compilation with Rust 1.67.0

bkchr:


tpwrules:
Thanks for the fix. Confirmed it works too.

PR #69, opened 2023-05-04
Fix Rust in kernel

### LINKED ISSUE #63: Only `llvmpipe` is detected as GPU backend on latest release
I build the latest release with the experimental mesa GPU driver, but only `llvmpipe` is detected as backend. Anyone seen this issue as well?

QuentinI:
Fixes #63 

Depends on #68 

This is certainly not how it _should_ be fixed. The bindgen version detection should ideally be fixed upstream, and the cross-references between `dev` and `out` should be properly addressed. Therefore, I'm not sure if you'll want to merge this. However, I wanted to submit anyway it so that other people could use it in the meantime

### DIFF
--- a/apple-silicon-support/packages/linux-asahi/default.nix
+++ b/apple-silicon-support/packages/linux-asahi/default.nix
@@ -74,6 +74,9 @@ let
       };
 
       kernelPatches = [
+        { name = "rust-bindgen-version";
+          patch = ./rust-bindgen-version.patch;
+        }
       ] ++ lib.optionals _4KBuild [
         # thanks to Sven Peter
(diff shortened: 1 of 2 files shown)


tpwrules:
Thank you very much for these fixes. I had to patch them up a bit to avoid issues when building without Rust. But everything works now.

QuentinI:
> I don't see any obvious issues with the approach.

I don't think it's technically wrong, just kind of unsatisfying. 

If I understood what's going on correctly it's the __kernel__ upstream who should ideally fix their script, because it is buggy and reads wrong version when run in a directory which contains a version.

And removing references is just annoying, since a lot of work is done to ensure everything is properly stripped at build stage, and I would prefer to solve the Rust part there as well. Unfortunately, that's a bit above my pay grade for now

PR #70, opened 2023-05-04
Remove IFD in kernel build

### LINKED ISSUE #51: Pure nix regex parser for the kernel config
The kernel package currently depends on an IFD (import from derivation) to parse the kernel config file.

The file is simple enough that it should be easily doable using pure nix with filter, map, regex match, etc.

This should be upstreamed to nixpkgs if resolved.

QuentinI:
Resolves #51

### DIFF
--- a/apple-silicon-support/packages/linux-asahi/default.nix
+++ b/apple-silicon-support/packages/linux-asahi/default.nix
@@ -10,16 +10,6 @@
 }:
 
 let
-  # TODO: use a pure nix regex parser instead of an IFD, and remove this workaround
-  localPkgs = if builtins ? currentSystem
-    then import (pkgs.path) {
-      crossSystem.system = builtins.currentSystem;
-      localSystem.system = builtins.currentSystem;
-    }
-    else pkgs;
-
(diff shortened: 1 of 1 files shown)


tpwrules:
Thank you very much for figuring this out. Is it still technically doing IFD though if there are kernelPatches?

QuentinI:
> Is it still technically doing IFD though if there are kernelPatches?

It's not importing a Nix expr, so as far as I understand it it's more lile a "Read From Derivation", which shouldn't be problematic

tpwrules:
I think it's still problematic because builds have to run before the evaluation can occur. I think the magnitude is much less than before but there is still IFD happening.

anna328p:
> https://github.com/tpwrules/nixos-apple-silicon/pull/70/files#diff-6db5403c64ac81d76097a9d10fbf02809552a396970b1a594d422515de2f0140R37-R43

It's possible to get rid of the IFD by getting rid of `writeFile` here and concatenating the config with the patches in Nix code. Instead of `readFile (writeFile foo)`, we just read `foo`, which isn't the output of a derivation.

I can PR this if needed.

PR #72, opened 2023-06-01
Update nixpkgs and rust-overlay

uuuvn:


tpwrules:
I would prefer to do these myself but an update is due soon with the release of 23.05.

uuuvn:
Sould i close this?

PR #73, opened 2023-06-01
Move away from wpa_supplicant to iwd

uuuvn:
It is better to use iwd instead of wpa_supplicant because the latter [does not support WPA3 on broadcom chips](https://www.reddit.com/r/AsahiLinux/comments/12igyoa/comment/jftvl3c) (which are installed on macs) and in general iwd is more modern, easy to use and maintained then wpa_supplicant.<br>
This PR is a draft because it must be merged at the same time as new release image is uploaded because otherwise it may confuse people who expect iwd to be availible in old installation ISO with wpa_supplicant.

### DIFF
--- a/docs/uefi-standalone.md
+++ b/docs/uefi-standalone.md
@@ -278,7 +278,21 @@ Some keyboard layouts are not detected correctly. On some devices, the \` key is
 
 Once you are happy with your initial configuration, you may install the system. This will have to download a large amount of data.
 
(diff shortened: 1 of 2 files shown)


tpwrules:
I don't think the disparaging comments are warranted and I would like them removed, but this seems reasonably reasonable. If you target the develop branch then it won't update the main guide at the listed URL and confuse people.

tpwrules:
The commit hash changed but I don't see that they are removed?

uuuvn:
Hmm, interesting. Let me check again

uuuvn:
image

uuuvn:
Looks like i forgot to do git add before git commit --amned, sorry

tpwrules:
Looks good and works, thanks.

PR #74, opened 2023-06-01
Update m1n1

uuuvn:


tpwrules:
Is it possible to simply grab the relevant commit(s) as a patch? The change described there should be small.

uuuvn:
Why patch old version when we can just upgrade?

tpwrules:
There have been several occasions where the Asahi folks are unhappy with people grabbing unreleased packages and sending them to users. I don't want to do it. If you want to upgrade you are welcome to and it's easy.

uuuvn:
It's tagged 1.2.8

uuuvn:
And AFAIK released, but in dev branch

uuuvn:
Then maybe separate branch (like asahi reference distro does) is the way to go?

uuuvn:
So people like me who use things like KVM or just want to have latest bleeding-edge version could use it instead of main?

tpwrules:
If you need to use it yourself you can create an overlay in your configuration to change m1n1's version.

uuuvn:
I already done so. The problem is that likely i'm not the only person who wants to eg use KVM. Why don't you think that a "dev" or "next" branch [like asahi reference distro does](https://github.com/AsahiLinux/PKGBUILDs/tree/dev) will be a great idea?

uuuvn:
@tpwrules by the way, yesterday asahi team released a new stable version of packages https://github.com/AsahiLinux/PKGBUILDs . Will you accept PR with versions from stable PKGBUILDs?

tpwrules:
Already been accepted.

PR #75, opened 2023-06-01
Update u-boot

uuuvn:
Used commit rev instead of tag because it contains various fixes, mainly usb-related crashes


tpwrules:
Not really comfortable grabbing untagged releases. I know the crashes are kind of annoying but they are not hard to resolve. Can the relevant commits just be added as patches?

uuuvn:
image

uuuvn:
Should i leave commit rev or switch to a tag?

PR #76, opened 2023-06-06
Package version bumps

natsukagami:
- Update m1n1 to 1.2.9
- Update linux-asahi to 6.3-6
- Update uboot-asahi to 2023.4.asahi2-1
- Update asahi-fwextract to 0.5.4
- Update mesa to 23.2.0_pre20230606-1

I also added the rest of them to the flake `packages` output, to make debugging easier...


apple-silicon-support/packages/linux-asahi/default.nix
```diff
@@ -7,7 +7,7 @@
 , linuxPackagesFor
 , _4KBuild ? false
 , withRust ? false
-, kernelPatches ? [ ]
+, _kernelPatches ? [ ]
```
natsukagami, line 10: Seemed like `callPackage` will fill this `kernelPatches` with... something. I get an `expected list got set` error before the change.

tpwrules:
Unfortunately it looks like the new U-Boot has problems booting a flash drive. Will try to work on a fix over the next week.

PR #80, opened 2023-06-25
add patch for rust 1.70

yu-re-ka:


bkchr:
Switching to `asahi-6.3-9` also fixes the compilation.

tpwrules:
Thanks again for the help!

PR #85, opened 2023-07-11
Update linux-asahi to 6.3-10

flokli:


tpwrules:
Thanks for alerting me there are new packages but I prefer to do the upgrade and release.

PR #86, opened 2023-07-12
Remove the Rust 1.70 patch

natsukagami:
I saw this was changed in the commit, so I thought it might be a mistake. The kernel rejects the patch on Rust 1.69 (latest nixos stable as of today) otherwise.

Build logs
```
applying patch /nix/store/1an7d69g4ig8dnnjxd1488j4l8c6ykn9-rust-1.70.patch
patching file rust/alloc/lib.rs
Reversed (or previously applied) patch detected!  Assume -R? [n] 
Apply anyway? [n] 
Skipping patch.
1 out of 1 hunk ignored -- saving rejects to file rust/alloc/lib.rs.rej
patching file rust/alloc/vec/into_iter.rs
Reversed (or previously applied) patch detected!  Assume -R? [n] 
Apply anyway? [n] 
Skipping patch.
5 out of 5 hunks ignored -- saving rejects to file rust/alloc/vec/into_iter.rs.rej
patching file scripts/Makefile.build
Reversed (or previously applied) patch detected!  Assume -R? [n] 
Apply anyway? [n] 
Skipping patch.
1 out of 1 hunk ignored -- saving rejects to file scripts/Makefile.build.rej
```

### DIFF
--- a/apple-silicon-support/packages/linux-asahi/default.nix
+++ b/apple-silicon-support/packages/linux-asahi/default.nix
@@ -116,15 +116,6 @@ let
           patch = ./rust_1_67_0.patch;
           reverse = true;
         }
-      ] ++ lib.optionals (rustOlder "1.70.0") [
-        { name = "rust-1.70.0";
-          patch = pkgs.fetchpatch {
-            name = "rust-1.70.patch";
-            url = "https://github.com/AsahiLinux/linux/commit/496a1b061691f01602aa63d2a8027cf3020d4bb8.diff";
-            hash = "sha256-RxOzCE22+yfZ3r/1sjbP8Cp+rRa56SJjHjwjTJxfIUI=";
-          };
-          reverse = true;
-        }
       ] ++ _kernelPatches;
 
       inherit configfile config;


natsukagami:
I'm not sure how it will do with Rust 1.70 though, so maybe it's not longer needed? https://github.com/tpwrules/nixos-apple-silicon/pull/80#issuecomment-1606232930 might be relevant.

tpwrules:
Maybe this patch should just be dropped completely? It should reverse the patch for Rust 1.69 since reverse = true. Did I put that in the wrong place somehow?

natsukagami:
I'm not sure if there is a `reverse = true` option for `kernel.manualConfig`.

https://github.com/NixOS/nixpkgs/blob/2de8efefb6ce7f5e4e75bdf57376a96555986841/pkgs/os-specific/linux/kernel/manual-config.nix#L31:L32

tpwrules:
Interesting. Is this patch backwards compatible? Does the kernel compile fine with Rust 1.69 if you cut that patch out completely?

I don't want to proclaim support for the stable release as that's an additional testing burden I can't really support but if the patch is unnecessary and removing it solves the problem of immediate failure I would be happy to remove it.

natsukagami:
It seems so. It builds fine and runs ok (I've been seeing some visual glitches on `swayfx` but I doubt it's the kernel's fault), on top of my 23.05 stable, on this branch (which uses Rust 1.69, so the patch should not kick in).

tpwrules:
Ok, I would be happy if you adjusted this PR to drop the patch entirely. Then it should support both current stable and unstable.

PR #91, opened 2023-08-16
Support flakes that pass `pkgs` to `nixosSystem`

Enzime:
When `pkgs` is passed to `nixosSystem` directly, the `nixpkgs.overlays` option doesn't do anything.

### DIFF
--- a/apple-silicon-support/modules/mesa/default.nix
+++ b/apple-silicon-support/modules/mesa/default.nix
@@ -7,7 +7,7 @@
     (lib.mkIf config.hardware.asahi.useExperimentalGPUDriver {
 
       # install the drivers
-      hardware.opengl.package = pkgs.mesa-asahi-edge.drivers;
+      hardware.opengl.package = config.hardware.asahi.pkgs.mesa-asahi-edge.drivers;
 
       # required for GPU kernel driver
       hardware.asahi.addEdgeKernelConfig = true;
@@ -17,7 +17,7 @@
       # without rebuilding them to avoid rebuilding the world.
       system.replaceRuntimeDependencies = [
         { original = pkgs.mesa;
-          replacement = pkgs.mesa-asahi-edge;
+          replacement = config.hardware.asahi.pkgs.mesa-asahi-edge;
         }
       ];
     })


tpwrules:
I was not actually aware that passing `pkgs` to `nixosSystem` was possible. Is that function documented anywhere?

I think the preferred method is adding a module which sets `nixpkgs.pkgs` to your `pkgs`. Then `nixpkgs.overlays` will work properly.

Enzime:
I can't find any documentation for it anywhere, but I know the three main module systems all support it (NixOS, nix-darwin and home-manager).

apple-silicon-support/modules/default.nix
```diff
@@ -21,6 +21,8 @@
             localSystem.system = cfg.pkgsSystem;
             overlays = [ cfg.overlay ];
           }
+        else if ! pkgs ? mesa-asahi-edge then
+          pkgs.extend cfg.overlay
```
Enzime, line 25: I'm happy to drop this part if you don't want to support people passing in `pkgs` directly.

apple-silicon-support/modules/mesa/default.nix
Enzime, line 1: I think the changes in this file should stay regardless, because I believe this module won't use the cross compiled versions when `hardware.asahi.pkgsSystem` is set.

apple-silicon-support/modules/default.nix
tpwrules: Two concerns:
* Not comfortable supporting something which breaks `nixpkgs.overlays`, that really should work
* This method of detecting that an un-overlayed set has been passed in seems fragile.

I would like this dropped.

apple-silicon-support/modules/mesa/default.nix
tpwrules: There is not a problem with this change per se, but I don't want to create the impression that I support cross-compiling Mesa. It should not be necessary to create a console installer image.

It may work, and that's cool, but I don't want that as a support consideration.

tpwrules:
Are you intending to cross-compile a full system?

apple-silicon-support/modules/default.nix
Enzime: Removed :+1:

Enzime:
The original reason behind why I made this PR was when I overrode `hardware.asahi.pkgs` with the overlay it still couldn't find `mesa`. I'm not intended to run a fully cross-compiled system at the moment, although it could be cool (combined with running a `x86_64-linux` builder locally like `darwin.linux-builder`)

Feel free to close this PR if you're not interested in this change at all

PR #92, opened 2023-08-21
Update uefi-standalone.md

aspauldingcode:
idevicerestore is a cross-platform command line utility designed specifically to restore and modify apple mobile "idevices," such as iPhone, and iPad. Not intended to restore Mac devices.

I've used idevice restore and never once seen the utility used for restoring another apple silicon mac, or intel mac.  Additionally, I haven't found any plans to support restoring mac firmware from idevicerestore or libimobiledevice. We could suggest such a feature to the idevicerestore developers, however.

To provide a more official recommendation, I provided the very tools I used to restore my m1 Macbook Air 2020 after breaking the recovery boot partition with asahi. These are official support guide and tools built from apple to restore apple silicon macs, even to restore macs which won't even boot into recovery.

### DIFF
--- a/docs/uefi-standalone.md
+++ b/docs/uefi-standalone.md
@@ -18,7 +18,8 @@ Perusing this guide might also be useful to users of other distros. Most of the 
 
 #### Warning
 
-Damage to the macOS recovery partitions or the partition table could result in the Mac becoming unbootable and loss of all data on the internal NVMe drive. In this circumstance, a suitable USB cable and another computer which can run [idevicerestore](https://github.com/libimobiledevice/idevicerestore) will be required to perform a DFU upgrade and restore normal operation. Backups are always wise.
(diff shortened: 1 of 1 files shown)


psanford:
Support for m1 macs was added to idevicerestore back in 2021. [Here's the initial PR](https://github.com/libimobiledevice/idevicerestore/pull/406).

idevicerestore is also listed a few times on the official asahi wiki as an option for recovery:
- https://github.com/AsahiLinux/docs/wiki/Introduction-to-Apple-Silicon
- https://github.com/AsahiLinux/docs/wiki/Partitioning-Cheatsheet
- and an example of running it: https://tg.st/u/idevicerestore_quickstart.txt

aspauldingcode:
@psanford, great news. Didn't know you could do that. 
Maybe we could provide those 3 links as well as keep the idevicerestore links instead of the ones I just added for apple configurator 2?

tpwrules:
I do not see that this is necessary. A complete and tested guide to using `idevicerestore` is provided [later on in the document](https://github.com/tpwrules/nixos-apple-silicon/blob/main/docs/uefi-standalone.md#recovering-from-boot-failure-with-idevicerestore). Maybe I will update the hyperlink to point there in case others overlook it.

PR #93, opened 2023-08-22
Add asahi-battery-threshold

Enzime:
`asahi-battery-threshold` is small daemon that makes it possible to set a charging threshold for laptops running Asahi Linux.

https://github.com/PaddiM8/asahi-battery-threshold

### DIFF
--- a/apple-silicon-support/modules/battery/default.nix
+++ b/apple-silicon-support/modules/battery/default.nix
@@ -1,0 +1,39 @@
+{ config, lib, ... }:
+{
+  config = let
+    cfg = config.services.asahi-battery-threshold;
+    pkgs' = config.hardware.asahi.pkgs;
+  in {
+    environment.etc."asahi-battery-threshold.conf".text = ''
+      start_charging_threshold = ${toString cfg.startCharging}
+      stop_charging_threshold = ${toString cfg.stopCharging}
+    '';
+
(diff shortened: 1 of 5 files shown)


Enzime:
Closing as this is natively supported by Asahi Linux https://twitter.com/AsahiLinux/status/1638954458821578752

Lucus16:
That tweet mentions it doesn't persist across reboots yet. And besides, I'd rather configure it in a NixOS module, so I'd appreciate it if this PR is reopened. I do think it's preferable to use udev rules over spawning a dedicated systemd service.

Enzime:
An example udev rule to make it persist across reboots is here:

https://github.com/Enzime/dotfiles-nix/blob/%F0%9F%94%A5/hosts/hermes/configuration.nix#L14-L16

leira:
@Enzime's link has stopped working. The udev lines are in [this revision](https://github.com/Enzime/dotfiles-nix/blob/6a99593b578b1cc0e147cc0e1d541475f13c338f/hosts/hermes/configuration.nix). Copy the lines below for future reference:

```
  services.udev.extraRules = ''
    KERNEL=="macsmc-battery", SUBSYSTEM=="power_supply", ATTR{charge_control_end_threshold}="80", ATTR{charge_control_start_threshold}="70"
  '';
```

PR #94, opened 2023-08-22
peripheral-firmware: extract on boot by default

Enzime:
This improves the situation for flake users, as previously they had to either add the non-redistributable firmware to their repositories or go without the firmware.

This change also brings the ISO into line with the regular NixOS modules as storing the peripheral firmware in the Nix Store is no longer a requirement.

### DIFF
--- a/apple-silicon-support/modules/peripheral-firmware/default.nix
+++ b/apple-silicon-support/modules/peripheral-firmware/default.nix
@@ -1,44 +1,71 @@
 { config, pkgs, lib, ... }:
 {
-  config = lib.mkIf config.hardware.asahi.enable {
-    assertions = lib.mkIf config.hardware.asahi.extractPeripheralFirmware [
-      { assertion = config.hardware.asahi.peripheralFirmwareDirectory != null;
+  imports = [
+    (lib.mkRemovedOptionModule [ "hardware" "asahi" "extractPeripheralFirmware" ] "This option is no longer necessary as the peripheral firmware will be extracted every boot automatically.")
+  ];
+
+  config = let
+    inherit (lib) mkIf;
+    cfg = config.hardware.asahi;
+    pkgs' = cfg.pkgs;
+  in mkIf config.hardware.asahi.enable {
+    assertions = mkIf cfg.usePeripheralFirmwareFromNixStore [
+      { assertion = cfg.peripheralFirmwareDirectory != null;
         message = ''
-          Asahi peripheral firmware extraction is enabled but the firmware
-          location appears incorrect.
+          Managing the peripheral firmware in the Nix Store is enabled
+          but the firmware location appears incorrect.
         '';
       }
     ];
 
-    hardware.firmware = let
-      pkgs' = config.hardware.asahi.pkgs;
-    in
-      lib.mkIf ((config.hardware.asahi.peripheralFirmwareDirectory != null)
-          && config.hardware.asahi.extractPeripheralFirmware) [
-        (pkgs.stdenv.mkDerivation {
-          name = "asahi-peripheral-firmware";
+    hardware.firmware = mkIf (cfg.peripheralFirmwareDirectory != null && cfg.usePeripheralFirmwareFromNixStore) [
+      (pkgs.stdenv.mkDerivation {
+        name = "asahi-peripheral-firmware";
 
-          nativeBuildInputs = [ pkgs'.asahi-fwextract pkgs.cpio ];
+        nativeBuildInputs = [ pkgs'.asahi-fwextract pkgs.cpio ];
 
-          buildCommand = ''
-            mkdir extracted
-            asahi-fwextract ${config.hardware.asahi.peripheralFirmwareDirectory} extracted
+        buildCommand = ''
+          mkdir extracted
+          asahi-fwextract ${cfg.peripheralFirmwareDirectory} extracted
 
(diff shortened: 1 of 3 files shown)


tpwrules:
I do not think I agree with this change at this time.

1. It exposes us to race conditions between drivers which need the firmware and the extraction of it.
2. It means firmware is less controlled by the generational mechanism which could create incompatibilities in updates. Admittedly this is not exceptionally well defined by default but these changes would make rolling back a firmware ESP file update through the bootloader menu impossible.
3. It brings Python and a bunch of libraries into the system closure.

Let's revisit this once the mechanism by which firmware updates occur is defined. I will give some more thought to the flakes case in a bit.

MatthewCroughan:
Is it possible we could have this available as a module option, off by default? I would like to use it, but would have to maintain my own fork of nixos-apple-silicon, or use @Enzime's branch if it's not maintained here.

nrabulinski:
@tpwrules would you be more willing to accept this change if the extraction happened during stage-1 boot instead of stage-2? Then to my knowledge there should be no race. Admittedly there's still the issue of not having the firmware tied to the generation, but if we moved firmware initialization/extraction from this PR to stage-1, we could also utilize something like agenix for encrypting the firmware and then decrypting it during stage-1, so if someone wants to have their config public, they still could. WDYT?

tpwrules:
Asahi Linux has not really changed on this aspect unfortunately, and I still think this PR is an awkward middle. I don't think a stage one hugely improves the issues I raised earlier, and there are some more issues like firmware being bigger and slower to extract (lengthening boot), and limited EFI partition space meaning large initrds are a big problem and I suspect an entire Python installation is bigger than the firmware.

It would be extremely awesome, and probably an unfeasible amount of work, if we could write a derivation that downloads the macOS installer ISO and extracts the firmware much like the Apple Silicon installer does. This could at its base be done with ranged requests to avoid downloading whole dozen gigabytes, and there might be Python libraries now that let us mount the images to copy the files out. I don't recall exactly how the installer does it but it has some shortcuts here we could reuse and improve upon as the files we need are relatively small.

But the more pragmatic solution I think is to simply use `boot.initrd.prepend` or similar (maybe it needs to be an adjustment to the kernel command line; I have not looked at this in a while) to load the `firmware.cpio` archive like other distros do, combined with a NixOS option to use this mode. It would need instructions, a tool, or perhaps a systemd service to re-run the extractor and update that archive but that is updated pretty rarely still. It would also still have downside 2, but again that's not a huge issue in practice it has turned out. I would happily accept such a PR if you could prepare one or modify this one; I know this has been an issue for a long time now.

Enzime:
Someone has already done that here with Nix even: https://github.com/JJJollyjim/firmware-abomination

I no longer use Asahi so I don't have much interest in updating this PR

GavBog:
@tpwrules I've been working on a tool, https://github.com/GavBog/aea-tools, that addresses the firmware extraction challenges you mentioned here. It uses the method you mentioned, using range requests to extract the firmware from the IPSWs Apple provides.

I implemented a Proof of Concept in examples/asahi_firmware for extracting all the firmware files. This example uses more of a "spray and pray" method downloading all firmware instead of just the files needed for boot on a specific model (could be potentially be improved). I also had to filter out some of the M2 wifi firmware since it errored with asahi_fwextract and I haven't gotten around to looking at a fix yet (I'd need someone with a M2 mac to help test).

It should also be noted that aea-tools is built for extracting from the AEA (Apple Encrypted Archive) format used to wrap the RootFS in macOS 15+ and will not work with prior versions that don't use this wrapping. The asahi team tests with version 13.5 so it is diverging from that.

waltmck:
@GavBog That's awesome! It shouldn't be necessary to extract from AEA until we implement M4 support (since that is not supported by MacOS 14), but this work will absolutely be needed then.

A while ago I was trying to get this working. I managed to reproduce the [firmware-abomination offsets generation](https://github.com/JJJollyjim/firmware-abomination) experiment on up-to-date firmware with [some scripts](https://github.com/waltmck/asahi-firmware-download), and the only reason I couldn't write a derivation to generate the offsets (which could then be cached) is that I needed to FUSE mount an APFS image which isn't possible in a Nix sandbox.

There is additional firmware that is needed---my scripts downloaded the `firmware` directory which contains bluetooth and wifi firmware. However, `all_firmware.tar.gz` contains three other things: `._fud_firmware`, `appleh13camerad`, and the `fud_firmware` directory. We need to reimplement [what asahi-fwextract does](https://github.com/UbuntuAsahi/asahi-fwextract/blob/072e02a590a8bfbea40399783a8566bb7d072b39/src/stub.py#L343) in order to write a derivation that generates `all_firmware.tar.gz` in the same format.

GavBog:
@waltmck fud_firmware only contains one file (and a symlink to that file) `Firmware/J313_InputDevice.im4p` which is found in the IPSW (not encrypted). appleh13camerad is in the RootFS dmg.aea at `usr/libexec/appleh13camerad`. My example extracts all 3... It should also be possible to use the same method for pre-macOS 15 just omit passing it through aea-tools. 

Having a tool find the offsets on the fly is cleaner (and more legally sound) than storing the offsets/sync points like firmware-abomination does.

Edit: It seems I actually ran into your [asahi-firmware-download](https://github.com/waltmck/asahi-firmware-download) repo before! I'm one of its 3 stars :smile:

PR #95, opened 2023-08-22
WIP: Update linux and mesa

Lucus16:
The build of the kernel still fails with `sed: no input files`, probably in ~`postPatch`~ `postFixup`.
Because `$dev/lib/modules/6.4.0-asahi/build/vmlinux` does not exist.

PR #96, opened 2023-08-23
m1n1: 1.2.9 -> 1.3.3

Enzime:
This fixes brightness control on a M1 MacBook Air and possibly other devices.

Based on https://github.com/AsahiLinux/PKGBUILDs/commit/f96d51311473b2e6073fbc003d0838f299b1ec6b


Enzime:
Running the newer m1n1 without updating the rest of the system seems to have made the system laggier, it probably should be bumped with #95

PR #102, opened 2023-09-13
alsa-ucm-conf-asahi: Add support for audio confs

IonAgorria:
This adds the package so that headphones volume work by overriding the upstream alsa-ucm-conf and copying the extra asahi ucm2 confs into destination folder

Not entirely sure how package should be added for default installation in modules

### DIFF
--- a/apple-silicon-support/packages/alsa-ucm-conf-asahi/default.nix
+++ b/apple-silicon-support/packages/alsa-ucm-conf-asahi/default.nix
@@ -1,0 +1,19 @@
+{ lib
+, fetchFromGitHub
+, alsa-ucm-conf }:
+
+(alsa-ucm-conf.overrideAttrs (oldAttrs: rec {
+  version = "3";
+
+  src_asahi = fetchFromGitHub {
+    # tracking: https://github.com/AsahiLinux/PKGBUILDs/blob/main/alsa-ucm-conf-asahi/PKGBUILD
+    owner = "AsahiLinux";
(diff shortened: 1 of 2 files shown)


apple-silicon-support/packages/alsa-ucm-conf-asahi/default.nix
```diff
@@ -0,0 +1,19 @@
+{ lib
+, fetchFromGitHub
+, alsa-ucm-conf }:
+
+(alsa-ucm-conf.overrideAttrs (oldAttrs: rec {
+  version = "1";
```
ivabus, line 6: There is version "3" already

yusefnapora:
Hey, just wanted to chime in that I ended up doing basically the same thing on my install, but the override causes everything that links to `alsa-lib` to be rebuilt from source, which ended up pulling in about 5GB of stuff on my machine and takes quite a while to build.

As an alternative to overriding the stock `alsa-config-ucm` package, you could set the `ALSA_CONFIG_UCM2` environment variable to the `ucm2` directory in your new combined package, using something like this:

```nix
let

environment.variables = {
  ALSA_CONFIG_UCM2 = "${new-ucm-conf-package-from-this-pr}/share/alsa/ucm2";
};
```

The `ALSA_CONFIG_UCM2` variable isn't documented especially well; I only stumbled on it in [this old nixos issue](https://github.com/NixOS/nixpkgs/issues/95243). But, I tried it out and can confirm that alsa will load UCM files from there if the var is set.

PR #109, opened 2023-10-17
update linux-asahi, m1n1

yu-re-ka:


zzywysm (commented):
Don't you have to remove the two patches later in this file, because their contents are now part of upstream u-boot and are redundant?

rowanG077:
@yu-re-ka What is still blocking this? I would like to install nixos on my M2 macbook pro which recently got support.

tpwrules:
I will have this done in the next couple days.

PR #112, opened 2023-10-22
Improve documentation

CodeWitchBella:
Thank you for awesome project with clear documentation.

There was one thing that I wish that it was mentioned in the docs, so here's PR adding that. I also hid the 4k page configuration section as it is currently not working so that people do not waste time on trying to set it up.


kdb424:
I ran into exactly this while setting it up just about the same time the PR was opened. It's worth leaving the note, but mentioning that it may be broken. I simply commented it out while it was broken, but can enable it later once it's working. Better to at least document it.

karlskewes:
Plus one to adding `iwd` recommendation. I did the install loop twice. :)

PR #113, opened 2023-10-30
improve kernel configuration

zzywysm:
Include some no-brainer improvements to the Asahi kernel config for NixOS.


tpwrules:
Please format the diff to be cleaner. I can't tell what's changed and what's just an artifact of `make config`.

zzywysm:
> Please format the diff to be cleaner. I can't tell what's changed and what's just an artifact of `make config`.

I agree with your request in spirit, however please consider that 

> Please format the diff to be cleaner. I can't tell what's changed and what's just an artifact of `make config`.

I agree with your request in spirit (readability is good), however, please consider that over the long-term it will be much easier to only commit post-`make oldconfig`-processed config files.  This way things will always be in a consistent order as dictated by the Kconfig system, even if it's not particularly readable.

The temptation to tack on a couple quick entries at the end is large (including with me).  But as someone who has metaphorically burned themselves by hand-editing kernel config files, I get a lot of comfort knowing the config has at a bare minimum been validated and updated by going through Kconfig.

tpwrules:
My intent is to reduce burden on myself as maintainer, and making the file an exact clone of the upstream Asahi one, plus a very small and defined set of changes, accomplishes that. The Nix infrastructure does process the config file through `make config` and I check that the additional options are properly set. I admit I don't know enough about kernel config to decide whether or not any particular change is a "no-brainer" or not, so minimizing the set of questions is important to me.

If you intend to make more significant changes than "minimal set of options needed to make a NixOS feature work properly as users expect" than they will probably not be accepted. Otherwise, please explain those minimal sets of options and adjust the PR to add them to the end of the file like the others.

zzywysm:
> My intent is to reduce burden on myself as maintainer, and making the file an exact clone of the upstream Asahi one, plus a very small and defined set of changes, accomplishes that. 

But which upstream Asahi kernel to use?  The Fedora Asahi kernel config, which is suboptimal for NixOS because it's intended to support all of the possible ARM64 platforms, which includes a bunch of completely irrelevant stuff like Rockchip, etc.?

https://copr-dist-git.fedorainfracloud.org/cgit/@asahi/kernel/kernel.git/tree/kernel-aarch64-16k-fedora.config

Or the current choice, the ALARM Asahi kernel config.  This config is much more targeted to Apple Silicon, but Hector has gone on and on about how his days as an ALARM maintainer are short-lived.  So how much longer will you be able to rely on it?

You say you don't know much about kernel config, and want to ease your maintenance burden.  You are forgetting another option, which is to delegate the kernel config maintenance to me, instead of carrying the world on your shoulders.

tpwrules:
I'm not sure how much longer it will last, but if that's the case then maybe it's better to switch over to the stock NixOS kernel config generator. Builds will be longer but easier to reason about.

I still can't identify what changes you want to make.

PR #114, opened 2023-10-30
update m1n1 and asahi-fwextract

zzywysm:
These are the most recent versions from the Asahi project.

### DIFF
--- a/apple-silicon-support/packages/asahi-fwextract/default.nix
+++ b/apple-silicon-support/packages/asahi-fwextract/default.nix
@@ -8,14 +8,14 @@
 
 python3.pkgs.buildPythonApplication rec {
   pname = "asahi-fwextract";
-  version = "0.6.9";
+  version = "0.6.14";
 
   # tracking version: https://github.com/AsahiLinux/PKGBUILDs/blob/main/asahi-fwextract/PKGBUILD
   src = fetchFromGitHub {
     owner = "AsahiLinux";
     repo = "asahi-installer";
     rev = "v${version}";
-    hash = "sha256-MkNi4EBgT4gfev/yWqYyw5HZxewj6XTfb8na+eI2iVo=";
+    hash = "sha256-tGfL6SLzYbO0qC6oM3W9YnyU2z2yvPGMjKV4MRAk49o=";
   };
 
   postPatch = ''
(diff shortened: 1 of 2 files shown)


tpwrules:
As a rule we are only using published versions. Do these solve some specific problem?

zzywysm:
> As a rule we are only using published versions. Do these solve some specific problem?

m1n1 1.4.3 includes fixes that hopefully allow Asahi Linux to finally run on the Mac minis with M2 Pro processors.

As for asahi-fwextract, it seems like a good thing to be ahead of the curve on.

rowanG077:
I disagree with this. The Asahi team has been clear time and time again that you really shouldn't follow their dev branch. It's unvalidated and stuff might break at any time or never work in the first place.

zzywysm:
> The Asahi team has been clear time and time again that you really shouldn't follow their dev branch. It's unvalidated and stuff might break at any time or never work in the first place.

This would be a more worrying concern if:
1) m1n1 had release branches that they tagged releases from.  Instead, all tagged releases are tagged straight from the main "dev branch".
2) This is an actual tagged release, 1.4.3, not some random snapshot

rowanG077:
The m1n1 version to be used is tagger [here](https://github.com/AsahiLinux/asahi-installer)

zzywysm:
As @rowanG077 just mentioned (but appears to have deleted since I received the email notification), the current `asahi-installer` code is shipping m1n1 version 1.4.3, which is what I'm proposing updating `nixos-apple-silicon` to.

https://github.com/AsahiLinux/asahi-installer/commit/41edadfc868137360ae2d2e519be8d8cea9cba14

tpwrules:
I checked and what users actually get is still v1.4.2. Your link is still a development commit. I define "released to users" and suitable for inclusion as what's available [here](https://github.com/AsahiLinux/PKGBUILDs/), though that may need to change as the Asahi project prioritizes Fedora.

I'm not sure how exactly the Asahi project prefers to work, but there's a clear distinction between arbitrary commit, tagged release, pushed to users in the dev repos, and pushed to users in the stable repos. I intend to stay on that last step, I don't have the bandwidth or machines to support anything else.

One of the most beautiful things about Nix is how easy it is to fiddle with the versions if you need to. If you have a machine where this is needed to work properly, then by all means make the change and run it. But I don't want to (in whatever sense I actually can) officially promote it.

I also would welcome PRs which help to solve issues faced in upcoming releases so that I'm not caught with an unexpected problem. Of course those would not be merged until the actual release, and in any case I don't see that that applies here.

PR #117, opened 2023-11-10
rust 1.73.0 patch

yu-re-ka:
Needed for rustc 1.73.0 which is currently in staging-next.

The patch is authored by me. This time around the patch is a bit more involved/hacky. An attempt at an explanation:
rust-next already has a patch for 1.73.0, but rust-next does not contain the driver module of the kernel crate, because that is not yet upstreamed to mainline.
The const_impl unstable feature is being reworked, and I could not make it work, got several compiler panics on the way.
Because the affected RawDeviceId trait is currently only implemented for one single struct, we can just manually insert this DeviceId struct in all the places where RawDeviceId was used as a generic, circumventing the use of const_trait.

### DIFF
--- a/apple-silicon-support/packages/linux-asahi/default.nix
+++ b/apple-silicon-support/packages/linux-asahi/default.nix
@@ -104,6 +104,10 @@ let
         { name = "rust-bindgen-version";
           patch = ./rust-bindgen-version.patch;
         }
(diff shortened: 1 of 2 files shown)


tpwrules:
Seems to build and work. Thanks a bunch.

PR #122, opened 2023-12-15
Speakers

yu-re-ka:
Requirements:

- A patched eleven secret herbs and spices kernel (or a certain kernel cmdline parameter)

### DIFF
--- a/apple-silicon-support/modules/sound/default.nix
+++ b/apple-silicon-support/modules/sound/default.nix
@@ -2,19 +2,33 @@
 
 {
   options.hardware.asahi = {
-    setupAlsaUcm = lib.mkOption {
+    setupAsahiSound = lib.mkOption {
       type = lib.types.bool;
       default = config.sound.enable;
       description = ''
-	Enable the Asahi-specific ALSA UCM2 configs in the global environment
-        so that headphone jack input and output work properly.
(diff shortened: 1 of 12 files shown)


apple-silicon-support/modules/sound/default.nix
```diff
@@ -2,19 +2,81 @@
...
       default = config.sound.enable;
       description = ''
-	Enable the Asahi-specific ALSA UCM2 configs in the global environment
-        so that headphone jack input and output work properly.
+        Set up the Asahi DSP components
       '';
     };
   };
 
   config = lib.mkIf config.hardware.asahi.setupAlsaUcm {
```
psanford, line 14: This line is still referencing `setupAlsaUcm` which you renamed above.

tpwrules:
Unfortunately I was unable to make this branch work on my machine (should be j316, 2021 M1 Max MBP 16"), nor cherry-pick the audio commits to make this work.

Please rebase and test on the latest release to ensure we are both on the same page. I will do the same, then we can discuss privately if that doesn't work.

I genuinely appreciate your work and pioneering on these matters and will try more myself, but you might have a faster solution.

tpwrules:
Thanks for the contribution. I've taken and cleaned this up and merged this stuff in the latest release.

PR #123, opened 2023-12-17
Use 115200n8 for serial console to match what early boot (m1n1, u-boot, etc.) is using and what the Central Scrutinizer expects.

sdier:
I spent a bit too much time debugging why Central Scrutinizer wasn't working immediately with NixOS, turns out it was this baud rate. Notably it differs from what the rest of boot uses, so I think it should match the defaults for those parts. Unfortunately Central Scrutinizer isn't configurable, so if we set everything else to 1.5M it would not work by default.


tpwrules:
This looks reasonable, thank you.

PR #127, opened 2023-12-21
linux-asahi: Add CONFIG_MUX_APPLE_DPXBAR

rowanG077:
Closes https://github.com/tpwrules/nixos-apple-silicon/issues/126

A hacky fix for now since you normally shouldn't manually write into it like this. But at the very least this makes it work until the config is properly synchronized. I tried to find the Fedora config but I don't know where it is.


tpwrules:
Thanks for the fix. I've half figured out their system but haven't completed implementation yet.

PR #128, opened 2023-12-22
flake support and systemd-initrd

mlyxshi:
flake users will fail to build `asahi-peripheral-firmware`

This PR add flake support and implement extract vendor firmware by systemd-initrd

Also, `curl https://alx.sh | sh` install `UEFI + UBoot` will install vendor firmware `/Volumes/EFI - NIXOS/vendorfw/firmware.cpio`. Therefore, I think we don't need `${asahi-fwextract}/bin/asahi-fwextract /tmp/.fwsetup/esp/asahi /tmp/.fwsetup/extracted` again.

### DIFF
--- a/apple-silicon-support/modules/peripheral-firmware/default.nix
+++ b/apple-silicon-support/modules/peripheral-firmware/default.nix
@@ -1,35 +1,34 @@
 { config, pkgs, lib, ... }:
 {
-  config = {
-    assertions = lib.mkIf config.hardware.asahi.extractPeripheralFirmware [
-      { assertion = config.hardware.asahi.peripheralFirmwareDirectory != null;
-        message = ''
-          Asahi peripheral firmware extraction is enabled but the firmware
-          location appears incorrect.
-        '';
-      }
-    ];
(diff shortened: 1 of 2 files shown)


tpwrules:
The problem here is that `/Volumes/EFI - NIXOS/vendorfw/firmware.cpio` can become out of date when `asahi-fwextract` gets upgraded. But so too can the `all_firmware.tar.gz` and the macOS stub partition. All three have happened so far.

My preferred solution would be to have a derivation download and extract the firmware from the Apple recovery image, but that's pretty gnarly and does not solve the third problem.

I'll have to think on accepting this for now. I'm not sure I like enabling systemd-initrd (need to research this more) or flakes by default in this PR.

cid-chan:
I'd like to chip in:
I think enabling systemd-initrd by default, should be done, with some sort of announcement.

Enabling systemd-initrd broke my boot-process on M1 macs, as I had a snippet in my activation-script that required tty-access to enter a password. (I fixed it by switching to systemd-ask-password)

While I think only a minority will face this specific issue, it would be good to know that this might pose an issue when updating.

PR #129, opened 2023-12-23
flake support

mlyxshi:
Compared to #128, `systemd-initrd` and `traditional initrd` are both supported. `flake` and `systemd-initrd` are non-mandatory.

I haven't tested the traditional initrd, I think it should work. 

>The problem here is that /Volumes/EFI - NIXOS/vendorfw/firmware.cpio can become out of date when asahi-fwextract gets upgraded. But so too can the all_firmware.tar.gz and the macOS stub partition. All three have happened so far.

>My preferred solution would be to have a derivation download and extract the firmware from the Apple recovery image, but that's pretty gnarly and does not solve the third problem.

This is a temporary solution for better flake user experience.

### DIFF
--- a/apple-silicon-support/modules/peripheral-firmware/default.nix
+++ b/apple-silicon-support/modules/peripheral-firmware/default.nix
@@ -1,69 +1,70 @@
 { config, pkgs, lib, ... }:
+let
+  pkgs' = config.hardware.asahi.pkgs;
+in
 {
-  config = {
-    assertions = lib.mkIf config.hardware.asahi.extractPeripheralFirmware [
-      { assertion = config.hardware.asahi.peripheralFirmwareDirectory != null;
-        message = ''
-          Asahi peripheral firmware extraction is enabled but the firmware
-          location appears incorrect.
-        '';
-      }
-    ];
+  options.hardware.asahi.extractPeripheralFirmware = lib.mkOption {
+    type = lib.types.bool;
+    default = true;
+    description = ''
+      Automatically extract the non-free non-redistributable peripheral
+      firmware necessary for features like Wi-Fi.
+    '';
+  };
 
-    hardware.firmware = let
-      pkgs' = config.hardware.asahi.pkgs;
-    in
-      lib.mkIf ((config.hardware.asahi.peripheralFirmwareDirectory != null)
-          && config.hardware.asahi.extractPeripheralFirmware) [
-        (pkgs.stdenv.mkDerivation {
-          name = "asahi-peripheral-firmware";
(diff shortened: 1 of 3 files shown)


zzywysm:
Some code review notes:

1.  You should be using `/lib/firmware/vendor`, not /lib/firmware .  This is documented in https://github.com/AsahiLinux/docs/wiki/Open-OS-Ecosystem-on-Apple-Silicon-Macs , and has been clearly supported for a long time thanks to https://github.com/AsahiLinux/linux/commit/07db737a08f4e8b0f679ddcf4332b34ff7a6e3bf
2.  The `firmware.cpio` file is intended to be loaded directly, as @tpwrules mentioned above.  Until we figure out how to do that, it seems much simpler to use the `firmware.tar` file.  This way, you can avoid creating and removing a `/tmp/.fwsetup/` directory, and instead do something like:

tar -xf /sysroot/boot/vendorfw/firmware.tar -C /sysroot/lib/firmware/vendor

zzywysm:
> 2\. The `firmware.cpio` file is intended to be loaded directly, as @tpwrules mentioned above.  Until we figure out how to do that...

I think I figured the broad outline out, but NixOS is going to fight us.

If you look at /boot/loader/entries/nixos-generation-*.conf, they follow this format:

```
title NixOS
version Generation 52 NixOS Uakari 24.05pre561409.54aac082a4d9 (Linux 6.6.0-asahi), built on 2023-12-29
linux /efi/nixos/v1a8mj7fhs60rjdwx8jxd33whx8b54ig-linux-6.6.0-asahi-Image.efi
initrd /efi/nixos/a2pdvxdrigzyfj3wfc3mfrisvpqal6m9-initrd-linux-6.6.0-asahi-initrd.efi
options init=/nix/store/pgmngxjb4lbagd9szv5gxb50hrmxgga8-nixos-system-greenrosetta-24.05pre561409.54aac082a4d9/init [BOOTARGS ELIDED]
machine-id [ELIDED]
```

The documentation of this format (https://uapi-group.org/specifications/specs/boot_loader_specification/ ) says that you can have _more than one `initrd` entry_, and that they are expected to be in `cpio` format.

The problem is that the nixpkgs code that generates these `.conf` files (https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/system/boot/loader/systemd-boot/systemd-boot-builder.py )  only ever expects to include one `initrd` entry.  And because it's written in Python and not Nix, this severely limits the code's extensibility.

I think unless we modify how NixOS interacts with systemd-boot, we should stick with `firmware.tar` and ignore `firmware.cpio`.

PR #151, opened 2024-01-27
Improvements jan2024

zzywysm:
* According to Ubuntu's documentation, choosing "none" as the IO scheduler is the best choice for fast random-access devices such as NVME. They don't get much faster than Apple's.

https://wiki.ubuntu.com/Kernel/Reference/IOSchedulers

* forcefully disable CMD_NFS to achieve the same security improvement as this Fedora Asahi Remix patch:

https://copr-dist-git.fedorainfracloud.org/cgit/@asahi/u-boot/uboot-tools.git/tree/0001-disable-NFS-support-by-default.patch?h=f39

* decrease the boot delay
* disable the submarine logo on boot because it looks goofy


tpwrules:
These are still opinionated changes without significant evidence to back up claims.

I don't see how the NFS disable applies to us; I don't think it's possible to get network in U-Boot on Apple Silicon anyway.

zzywysm:
I don't think these changes are particularly opinionated.  They are mostly intended to more closely match the state of NixOS itself and of the builds Fedora Asahi Remix ships.

1. The NixOS installer ISO attempts to disable IO schedulers by passing the `elevator=noop` boot argument to the kernel.  (This is an antiquated way to do so, and I believe modern kernels ignore the argument.  Nevertheless, it indicates that NixOS itself has the opinion that the IO schedulers should be disabled.)
2. By disabling NFS, we match how Fedora builds u-boot, and remove a source of uncertainty should we ever face difficulties with u-boot that require debugging.
3. Decreasing the u-boot boot delay also decreases the total boot time, which is a nice benefit for users.

zzywysm:
Some experimental data on IO schedulers on NVME drives:

https://www.phoronix.com/review/linux-56-nvme

PR #153, opened 2024-01-29
mesa-asahi-edge: update disk cache patch

zzywysm:
This patch hasn't been updated in five months, and the line numbers don't match the state of the code in the asahi-20231213 tag.

This new version of the patch is taken verbatim from nixpkgs HEAD.  Verified via manual inspection that the line numbers match the asahi-20231213 tag.


tpwrules:
I don't understand the problem this solves. If the patch still applies then there's no benefit to changing it.

PR #179, opened 2024-03-14
asahi-nvram: add asahi-nvram 0.2.1

lukaslihotzki:
This PR adds the asahi-nvram tool. The other tools ([asahi-bless](https://github.com/WhatAmISupposedToPutHere/asahi-nvram/tree/master/asahi-bless), [asahi-btsync](https://github.com/WhatAmISupposedToPutHere/asahi-nvram/tree/master/asahi-btsync), [asahi-wifisync](https://github.com/WhatAmISupposedToPutHere/asahi-nvram/tree/master/asahi-wifisync)) should be similar.

### DIFF
--- a/apple-silicon-support/packages/asahi-nvram/default.nix
+++ b/apple-silicon-support/packages/asahi-nvram/default.nix
@@ -1,0 +1,16 @@
+{ fetchCrate
+, rustPlatform
+}:
+
+rustPlatform.buildRustPackage rec {
+  pname = "asahi-nvram";
+  version = "0.2.1";
+
+  src = fetchCrate {
+    inherit pname version;
+    hash = "sha256-bFUFjHVTYj0eUmhijraOdeCvAt2UGX8+yyvooYN1Uo0=";
+  };
+
+  cargoHash = "sha256-WhySIQew8xxdwXLWkpvTYQZFiqCEPjEAjr7NVxfjDkU=";
+  cargoDepsName = pname;
+}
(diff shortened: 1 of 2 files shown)


tpwrules:
It is certainly not attractive that the primary documentation I can find on these tools is "May or may not result in you having to perform a dfu restore.".

I had a brief look though and it seems that they are available in Fedora but not installed by default. Of note, I don't think `asahi-wifisync` will do much because we do not ship iwd by default.

`asahi-bless` looks pretty useful though, and harder to break things with, I'd be fine with just doing that one.

alyssais:
Wouldn't it make more sense to put this in Nixpkgs?

tpwrules:
I would prefer that, yes.

lukaslihotzki:
Ok, I can try to contribute these packages to nixpkgs. After this, it would make sense to also upstream asahi-fwextract, bankstown-lv2, and speakersafetyd.

tpwrules:
I'm not so sure about asahi-fwexctract since it has to match closely with the other components.

flokli:
@lukaslihotzki did you start sending PRs already? Was looking for these tools today, and couldn't find them here or in nixpkgs.

lukaslihotzki:
I have now opened a PR in nixpkgs. I don't know if this will get merged there, but at least the branch can be used manually.

lukaslihotzki:
Now upstreamed by https://github.com/NixOS/nixpkgs/pull/302205

PR #183, opened 2024-04-01
modules/sound: select `asahi-audio` version conditionally

fx-chun:
`asahi-audio` is committing to support of a `compat/1.x` branch for Wireplumber versions below `0.5` until near the end of 2024: https://github.com/AsahiLinux/asahi-audio/commit/04e905abdb5f561126fcd29be809d5b3ff95b820

This PR splits the `asahi-audio` package into `1_x` and `2_x` variants, allowing us to select a version conditionally based on the Wireplumber version in `nixpkgs`. This allows users to continue tracking whatever version of `nixpkgs` independent from the Wireplumber version.

### DIFF
--- a/apple-silicon-support/modules/sound/default.nix
+++ b/apple-silicon-support/modules/sound/default.nix
@@ -60,6 +60,9 @@
       systemd.user.services.pipewire.environment.ALSA_CONFIG_UCM2 = config.environment.variables.ALSA_CONFIG_UCM2;
       systemd.user.services.wireplumber.environment.ALSA_CONFIG_UCM2 = config.environment.variables.ALSA_CONFIG_UCM2;
 
+      # wireplumber 0.5 looks for scripts in datadirs only
+      systemd.user.services.wireplumber.environment.XDG_DATA_DIRS = "${asahi-audio}/share";
+
       # enable speakersafetyd to protect speakers
       systemd.packages = lib.mkAssert lsp-plugins-is-safe
         "lsp-plugins is unpatched/outdated and speakers cannot be safely enabled"
--- a/apple-silicon-support/modules/sound/default.nix
+++ b/apple-silicon-support/modules/sound/default.nix
@@ -69,11 +69,23 @@
         [ pkgs.speakersafetyd ];
       services.udev.packages = [ pkgs.speakersafetyd ];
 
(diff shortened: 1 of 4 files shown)


flokli:
Can you elaborate where we'll need wireplumber < 0.5?

nixpkgs unstable is already using 0.5.1, and that'll be part of the upcoming NixOS release too. So in ~2 months from now there won't be a nixpkgs around with the pre-0.5 wireplumber version.

fx-chun:
I view 2~3 months as the minimum timeframe that we should probably support both Wireplumber major versions (time until next release + 1 month overlap); it's a significant period of time. Additional reasons are that I suspect there are going to be people who want/need to run on NixOS 23.11/Wireplumber 0.4 for a while & also want to respect upstream's decision to support both branches for that long.

If people think it's unnecessary to maintain support for both major versions I can create a standalone PR involving 0.5 only 🙂

fx-chun:
Tested with the latest Wireplumber master since 0.5.2 hasn't been released yet:

```
nixpkgs.overlays = lib.mkOrder 100 [
    (final: prev: {
      wireplumber = prev.wireplumber.overrideAttrs (old: rec {
        version = "0.5.2+git20240412";
        src = final.fetchFromGitLab {
          domain = "gitlab.freedesktop.org";
          owner = "pipewire";
          repo = "wireplumber";
          rev = "473e463c563f67ef6da72d63a9f8daecfa42f1e3";
          hash = "sha256-9tPz309rezFkiMABfNQ6ncQ5fZEDGjUbOgpRVeopTeU=";
        };
      });
    })
  ];
```

tpwrules:
I've been thinking about this and I would prefer to have only one version. Wireplumber 0.5.2 is in nixpkgs unstable now, so that should be good. I did test asahi-audio 1.8 and it sounded worse to me than 1.6. I need to test 2.0 still though.

I've had to leave behind the stable users in the latest kernel update. Let's give it a week or so for feedback and if there's not major issues, we'll just switch to 2.0 now. I don't think there's any critical features or fixes so users on older releases shouldn't be left behind.

fx-chun:
@tpwrules OK! I think that's fair. I just opened up #195 with the 2.x-exclusive changes.

PR #188, opened 2024-04-18
idevicerestore in master and nixpkgs-unstable are the same

okin:


tpwrules:
This whole section needs to be adjusted, as users don't really need to clone nixpkgs anymore either.

flokli:
The `idevicerestore` version in 23.11 is also the same. I sent an updated PR in https://github.com/tpwrules/nixos-apple-silicon/pull/201.

PR #191, opened 2024-04-25
Update to asahi linux kernel asahi-6.8.6-3

oliverbestmann:
Things seem to work on my M1 pro as far as I can see. I have not tested HDMI yet.

* Update to kernel version asahi-6.8.6-3 
* Adjust config options to match [fedora](https://gitlab.com/fedora-asahi/kernel-asahi/-/commits/fedora-asahi-6.8.7-pre/redhat/configs/fedora?ref_type=heads)
* Remove rust 1.76 and rust 1.77 patches, as they seem to be already applied in 6.8


apple-silicon-support/packages/linux-asahi/config
```diff
@@ -7843,3 +7843,4 @@ CONFIG_SPI_HID_APPLE_OF=m
 CONFIG_SPMI_APPLE=m
 CONFIG_VIDEO_APPLE_ISP=m
 CONFIG_SND_SOC_CS42L84=m
+CONFIG_APPLE_SIO=m
```
rowanG077, line 7846: Broken newline

tpwrules:
Thank you. Manually merged to clean up and fix some stuff: https://github.com/tpwrules/nixos-apple-silicon/commit/c501fa8f33020afdb3e7c97c29502cd24f46f78a

This does break stable users but we'll see if someone can contribute a patch. I'm not aware of huge improvements with this version anyway.

oliverbestmann:
Nice, thanks!

What do you mean with "breaks stable users"?

Regarding improvements: I observed a 10% score increase when running glmark2 after updating the kernel to 6.8 on my m1 pro compared to the 6.6 kernel. This might also have a positive effect on wayland and friends.

tpwrules:
The Rust stuff is no longer compatible with the compiler in stable nixpkgs so that support is broken. I tried reverting the patches but couldn't figure it out.

PR #203, opened 2024-05-25
mesa: fix cross compiling

RossComputerGuy:
This PR prevents weird behavior when cross compiling to other architectures and mesa is wanted.

### DIFF
--- a/apple-silicon-support/modules/mesa/default.nix
+++ b/apple-silicon-support/modules/mesa/default.nix
@@ -37,7 +37,8 @@
       # (and in a way compatible with pure evaluation)
       nixpkgs.overlays = [
         (final: prev: {
-          mesa = final.mesa-asahi-edge;
+          # prevent cross-built Mesas that might be evaluated using this config (e.g. Steam emulation via box64) from using the special Asahi Mesa
+          mesa = if prev.targetPlatform.isAarch64 then final.mesa-asahi-edge else prev.mesa;
         })
       ];
     })


tpwrules:
I do not understand the motivation here. I'm not sure I can support cross-compiling our Mesa, though I am a little surprised that it does not work.

If you need to make this conditional, it should suffice to check in your own configuration and set `hardware.asahi.useExperimentalGPUDriver` appropriately rather than modifying nixos-apple-silicon.

In any case, this modification here puzzles me as it keeps the kernel mode components enabled, but the driver can't work without the user mode components too.

RossComputerGuy:
> I do not understand the motivation here. I'm not sure I can support cross-compiling our Mesa, though I am a little surprised that it does not work.

This fixes using Mesa when it's an overlay and you're cross compiling for a different architecture. Without this, if I were to use pkgsCross.gnu64.mesa it would build Asahi Mesa instead of the x86 one. This PR prevents the Asahi Mesa to be used outside of aarch64.

RossComputerGuy:
> In any case, this modification here puzzles me as it keeps the kernel mode components enabled, but the driver can't work without the user mode components too.

I had more stuff in this PR without realizing how the module worked but forgot to change the title. Disregard the part that says this allows for disabling. This PR only allows for the Asahi Mesa to be used on aarch64.

tpwrules:
Oh okay, so you are saying if you are on a machine with nixos-apple-silicon and want to cross-build Mesa for an x86 machine? That makes sense now, but how would this overlay get into that version of nixpkgs? I am reasonably sure it will not if you do e.g. `nix-build -A pkgsCross.gnu64.mesa '<nixpkgs>'`.

I suppose if you have everything in the same config somehow? Why doesn't setting `hardware.asahi.enable` to false take care of that?

RossComputerGuy:
> Oh okay, so you are saying if you are on a machine with nixos-apple-silicon and want to cross-build Mesa for an x86 machine?

So I could play Steam games using box64.



> That makes sense now, but how would this overlay get into that version of nixpkgs?

Because I enabled the Steam module in NixOS.

> I suppose if you have everything in the same config somehow? Why doesn't setting `hardware.asahi.enable` to false take care of that?

Because I don't know what else that would disable.

tpwrules:
Okay, I see now, that makes sense. Thank you for explaining. I don't think that is an issue with the other options because presumably the Steam module doesn't use `hardware.opengl` for the cross build and `system.replaceRuntimeDependencies` won't have the same source Mesa.

Please add a comment summarizing the purpose like `# prevent cross-built Mesas that might run on this system under emulation (e.g. box64 and Steam) from using the special Asahi Mesa` and I will merge.

RossComputerGuy:
> Please add a comment summarizing the purpose like `# prevent cross-built Mesas that might run on this system under emulation (e.g. through Steam and box64) from using the special Asahi Mesa` and I will merge.

Will do when I am able to this evening. 👍

RossComputerGuy:
Done

apple-silicon-support/modules/mesa/default.nix
```diff
@@ -37,7 +37,8 @@
       # (and in a way compatible with pure evaluation)
       nixpkgs.overlays = [
         (final: prev: {
-          mesa = final.mesa-asahi-edge;
+          #  revent cross-built Mesas that might run on this system under emulation (e.g. through Steam and box64) from using the special Asahi Mesa
```
tpwrules, line 40: Copy-paste error and rephrasing, thanks again.

```suggestion
          # prevent cross-built Mesas that might be evaluated using this config (e.g. Steam emulation via box64) from using the special Asahi Mesa
```

tpwrules (changes requested):

apple-silicon-support/modules/mesa/default.nix
RossComputerGuy: Done :upside_down_face:

PR #213, opened 2024-07-08
mesa: Fix compilation with latest nixpkgs master

bkchr:
Requires the following upstream pr right now: https://github.com/NixOS/nixpkgs/pull/325449

### DIFF
--- a/apple-silicon-support/packages/mesa-asahi-edge/default.nix
+++ b/apple-silicon-support/packages/mesa-asahi-edge/default.nix
@@ -7,10 +7,8 @@
 
 (mesa.override {
   galliumDrivers = [ "swrast" "asahi" ];
-  vulkanDrivers = [ "swrast" ];
   enableGalliumNine = false;
-  # libclc and other OpenCL components are needed for geometry shader support on Apple Silicon
-  enableOpenCL = true;
+  vulkanDrivers = [ "swrast" ];
 }).overrideAttrs (oldAttrs: {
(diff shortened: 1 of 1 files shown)


ocfox:
Nice work!

damien-biasotto:
So according to the feedback you got on your PR on nixpkgs, do you think the recommended way would be to add `"-Dgallium-nine=disabled"` to the `mesonFlags` list in `packages/mesa-asahi-edge/default.nix`? 

Or is there more to it?

tpwrules:
I’m thinking the right solution might be to just vendor the whole mesa expression code instead of trying to override three ways at once. This would also help avoid problems caused by differences between stable and unstable. I don’t think mesa’s dependencies change that much but the expression is pretty volatile it seems.

damien-biasotto:
> I’m thinking the right solution might be to just vendor the whole mesa expression code instead of trying to override three ways at once. This would also help avoid problems caused by differences between stable and unstable. I don’t think mesa’s dependencies change that much but the expression is pretty volatile it seems.

I'll be keen to help you achieve this, but I'm a bit clueless how to achieve it. If you could give me some pointers, I would prepare a PR.

tpwrules:
Fixed in the latest release. Ended up going with the vendor solution.

bkchr:
Ty @tpwrules :) Vendoring sounds like a proper solution. I had some different change locally, but yours look good.

PR #226, opened 2024-08-05
Bumping asahi kernel to 6.9.12-2

### LINKED ISSUE #225: Can't connect to WiFi networks
After updating my 2021 MBP 14' I couldn't connect to any WiFI networks anymore. I eventually found [this](https://github.com/raspberrypi/linux/issues/6049#issuecomment-2259734713) workaround for a similar issue with Raspberry Pi, and adding `brcmfmac.feature_disable=0x82000` to `boot.kernelParams` solved the problem.

I don't know if this affects anyone else and whether it's something this project would want to handle at all and if yes, then what the preferred way would be (make it default, make it an optional toggle, just mention it in README), so I'm just opening this issue to hopefully save people who might run into similar issues some time, because I don't think I've seen any mention of this workaround specifically in connection to Asahi anywhere.

Please feel free to close this as the actual issue on my end is solved.

damien-biasotto:
Changing the kernel tag to the latest one (bundling the wifi driver fix)

Should fix #225.

### DIFF
--- a/apple-silicon-support/packages/linux-asahi/default.nix
+++ b/apple-silicon-support/packages/linux-asahi/default.nix
@@ -94,8 +94,8 @@ let
         # tracking: https://github.com/AsahiLinux/linux/tree/asahi-wip (w/ fedora verification)
         owner = "AsahiLinux";
         repo = "linux";
-        rev = "asahi-6.9.12-1";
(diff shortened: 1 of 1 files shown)


tpwrules:
Updated in the latest release, sorry for the delays.

PR #231, opened 2024-09-01
mesa: Upstream removed lib.platforms.mesaPlatforms

rowanG077:
This fixes the build with current nixos-unstable. 

See https://github.com/NixOS/nixpkgs/pull/338337


ibveecnk:
Thanks for the fix, I can confirm it works.

As I can see, the platforms are now contained in mesas `(mesa.meta) platforms`

https://github.com/alyssais/nixpkgs/blob/ef3692d0559c3ba8b682f07fd8d19669bbee4893/pkgs/development/libraries/mesa/common.nix#L32-L36

I am unsure but but isn't it possible to inherit those platforms in the `mesa-asahi-edge` platforms instead of having two sources of truth?

rowanG077:
The entire mesa package is vendored. Because of this I chose to not depend on the original at all. If @tpwrules wants to do that I can change it.

ibveecnk:
Thats a good point, let's see what @tpwrules says.

tpwrules:
This works and looks good. I would indeed prefer to duplicate the list rather than just change to inherit from a different place. Thank you!

PR #250, opened 2024-11-19
bump kernel and add a new option support fan control

marcusramberg:
- **linux-asahi: asahi-6.10.6-1 -> asahi-6.10.9.3**
- **modules/kernel: add option to enable hwmon fan control**

Tested this on my mac studio m1, being able to slow down the fan makes a huge noise difference.

### DIFF
--- a/apple-silicon-support/modules/kernel/default.nix
+++ b/apple-silicon-support/modules/kernel/default.nix
@@ -54,6 +54,10 @@
       "hid_generic"
     ];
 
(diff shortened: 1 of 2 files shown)


flokli (changes requested):

flokli:
There's been a few other kernel bumps since then (to 6.11 too), this could probably be re-rolled separately.

tpwrules:
Thanks for the effort, but I am not interested in so easily enabling an option so direly labeled. This should be easy for users to add to their own config if they so desired.

PR #252, opened 2024-11-22
systemd: only apply boot fix patch for 256.7

flokli:
That's the only released version accessible in nixpkgs that exposed this issue.

It's fixed in 256.8, which is already in nixpkgs staging.

If we unconditionally apply it for all later versions, the pach will fail to apply and due to this fail the build.


tpwrules:
Thanks! I expected this to happen but wanted to confirm and break people's builds instead of their boots.

Issue #257, opened 2024-12-06
Plymouth is not showing


bruhderek:
Can anyone confirm my issue with plymouth not showing up? Silent boot is also not working. Here's my config:
```nix
#boot.loader
    plymouth = {
      enable = true;
      theme = "rings";
      themePackages = with pkgs; [
        # By default we would install all themes
        (adi1090x-plymouth-themes.override {
          selected_themes = [ "rings" ];
        })
      ];
      extraConfig = ''
        ShowDelay=0
      '';
    };

    # Enable "Silent Boot"
    consoleLogLevel = 3;
    initrd = {
      verbose = false;
      # systemd.enable = true;
    };
    kernelParams = [
      "quiet"
      "loglevel=3"
      "udev.log-priority=3"
      "splash"
    ];
  ```
No matter what I do, I can't get plymouth to show :angry:

youwen5:
Latest revision fails to boot, systemd seems to be stuck at LUKS decryption. Plymouth also doesn't show over kernel logs before it crashes.

marcusramberg:
It fails to boot for me too. Unfortunately I’m also unable to get the boot menu to respond to keyboard input, so I’m a bit stuck on booting an older version as well.

tpwrules:
What system are you using? Please try a USB keyboard.

marcusramberg:
> What system are you using? Please try a USB keyboard.

It's a m1 studio, and I'm using a usb-a keyboard. I've also tried connecting a usb-c keyboard, and connecting a usb-a keyboard via a usb-c hub, with no success. gonna see if I can get a usb stick boot to work now.

tpwrules:
I think only some USB ports work.

marcusramberg:
> I think only some USB ports work.

That's a great hint! The rightmost usb-c port on the back worked. Back in business now.

tpwrules:
Glad to hear it. Please file an issue, including your machine type and any journalctl logs you can get from a failed boot, or anything visible on the monitor.

marcusramberg:
Mine is failing after waiting on a disk that doesn't mount, I'll reboot back and get a screenshot.

marcusramberg:
![image](https://github.com/user-attachments/assets/7d87aeb9-55f3-4f6e-943f-fd67235aa24e)
![image](https://github.com/user-attachments/assets/dce06f3d-b8ad-4242-a799-76193a11dc6e)
As mentioned this is a m1 studio max.

[...]

PR #273, opened 2025-02-08
Remove serial console commandline

LexNastin:
**What?**
This PR removes the serial console parameter from the kernel commandline

**Why?**
1. This causes plymouth to break. I don't know why it is so, but it is so. This PR would fix issue #257 (which I was also struggling with)
2. It's a little unnecessary. It's not core to getting NixOS to boot and work well (which I assume is the goal of this repo), and if a user wants the serial functionality, they can always add it back in their own NixOS configuration (much easier to add after the fact than trying to apply a patch/overlay something onto this repo to remove the kernel parameter without forking)

### DIFF
--- a/apple-silicon-support/modules/kernel/default.nix
+++ b/apple-silicon-support/modules/kernel/default.nix
@@ -56,7 +56,6 @@
 
     boot.kernelParams = [
       "earlycon"
-      "console=ttySAC0,115200n8"
       "console=tty0"
       "boot.shell_on_fail"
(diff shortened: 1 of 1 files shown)


tpwrules:
Thanks for the fix. As I recall this is necessary for proxy debugging with m1n1 but that is a rare use case and support has degraded anyway.

In the future we might be able to use the kernel command line option `plymouth.ignore-serial-consoles`?

flokli:
We should not be setting any console in the kernel cmdline. This is coming from the device tree normally.

PR #281, opened 2025-03-12
linux-asahi: 6.12.12-asahi -> 6.13.5-asahi

flokli:
The former doesn't build with rustc 1.85, which is what's landed in nixpkgs.


tpwrules:
Thank you. The config options can be cleaned up a bit but I will do this in a follow-up commit.

PR #284, opened 2025-03-16
use mesa instead of the deprecated mesa.drivers

marcin-serwin:
After https://github.com/NixOS/nixpkgs/pull/387292 mesa.drivers became deprecated and now points to the nixpkgs mesa which means that asahi-mesa is not used even if experimental driver is enabled.

### DIFF
--- a/apple-silicon-support/modules/mesa/default.nix
+++ b/apple-silicon-support/modules/mesa/default.nix
@@ -17,7 +17,7 @@
     }
     (lib.mkIf config.hardware.asahi.useExperimentalGPUDriver {
       # install the Asahi Mesa version
-      hardware.graphics.package = config.hardware.asahi.pkgs.mesa-asahi-edge.drivers;
+      hardware.graphics.package = config.hardware.asahi.pkgs.mesa-asahi-edge;
       # required for in-kernel GPU driver
       hardware.asahi.withRust = true;
     })
--- a/apple-silicon-support/packages/mesa-asahi-edge/default.nix
+++ b/apple-silicon-support/packages/mesa-asahi-edge/default.nix
@@ -37,6 +37,7 @@
   # replace patches with ones tweaked slightly to apply to this version
   patches = [
     ./opencl.patch
+    ./system-gbm.patch
   ];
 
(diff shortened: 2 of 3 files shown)


schphe:
you also need to apply new patches [like this](https://github.com/tpwrules/nixos-apple-silicon/compare/main...schphe:nixos-apple-silicon:main), otherwise the build will fail

Lederstrumpf:
> you also need to apply new patches [like this](https://github.com/marcin-serwin/nixos-apple-silicon/compare/push-nwvktpxoswts...schphe:nixos-apple-silicon:mesa-changes), otherwise the build will fail

This works for me - thanks!
For anyone else stumbling into this - gitlab.freedesktop is undergoing planned outage until end of the week (22 March), so fetching asahi-mesa currently 503's.

Here's a temporary override atop @schphe's patches using https://sources.voidlinux.org/mesa-asahi-25.1.0%2B20250221/mesa-asahi-20250221.tar.gz:
https://github.com/Lederstrumpf/nixos-apple-silicon/commit/282fb534763273609fa9d71acee95761d743bf10

flokli:
I got bit by this today, the workaround in nixpkgs sending the warning of `mesa.drivers` being deprecated actually silently falls back to mainline mesa, causing everyone with a more recent nixpkgs to not use asahi mesa at all.

@tpwrules can this be merged?

montchr:
This is unfortunately not working for me.  When I try launching Hyprland via regreet, I run into the error message about missing gbm, as described in this thread: https://discourse.nixos.org/t/mesa-loader-failed-to-find-dri-gdm-so/61015

montchr:
@yuyuyureka in this case, yes.  Once I updated my flake inputs to latest, that error no longer appears.  But this brings me back to the place where I have been stuck for weeks: Hyprland cannot start because, for some reason not apparent in logs, `WAYLAND_DISPLAY` is unset.  This leads to UWSM bailing out and ending the graphical session, where I cannot even switch to another tty.  Lately, I have been using https://github.com/flokli/nixos-apple-silicon/tree/wip (containing several of these open PRs) but that has not helped.

PR #287, opened 2025-03-21
update asahi-audio, alsa-ucm-conf-asahi, and add triforce-lv2 to sound module for mic support

kitten:
## Summary

These are changes manually copied over from an `edge` branch on a fork of this repo I'm playing around with. As per today's announcement, these audio updates have landed in the next Asahi Fedora Remix version and are relevant for `asahi-6.13.7-4` onwards.

> [!WARNING]
> This will only work if the Kernel is updated to track at least `6.13.7-4`!
> `-4` specifically disables flags that block out mic support added via `aop-audio`. This isn't in this PR to keep this PR focused on what it specifically updates (audio/mic updates)

Depends on https://github.com/tpwrules/nixos-apple-silicon/pull/289 to bump nixpkgs to include triforce-lv2.

## Set of changes

- Update asahi-audio to 3.1 (tracking)
- Add missing `passthru.requiredLv2Packages` to `asahi-audio` since it'll help with the sound module in the future
- Update alsa-ucm-conf-asahi to v7 (tracking)
- Add new `triforce-lv2` package to overlay
- Update sound module with new `triforce-lv2` lv2 package

### DIFF
--- a/apple-silicon-support/modules/sound/default.nix
+++ b/apple-silicon-support/modules/sound/default.nix
@@ -22,7 +22,6 @@
     lsp-plugins-is-safe = (pkgs.lib.versionAtLeast lsp-plugins.version "1.2.14");
 
     lv2Packages = with pkgs; [ lsp-plugins bankstown-lv2 triforce-lv2 ];
-    lv2Path = lib.makeSearchPath "lib/lv2" ls2Packages;
   in lib.mkIf (cfg.setupAsahiSound && cfg.enable) (lib.mkMerge [
     {
       # can't be used by Asahi sound infrastructure
--- a/apple-silicon-support/modules/sound/default.nix
+++ b/apple-silicon-support/modules/sound/default.nix
@@ -21,7 +21,8 @@
 
     lsp-plugins-is-safe = (pkgs.lib.versionAtLeast lsp-plugins.version "1.2.14");
 
-    lv2Path = lib.makeSearchPath "lib/lv2" [ lsp-plugins pkgs.bankstown-lv2 ];
+    lv2Packages = with pkgs; [ lsp-plugins bankstown-lv2 triforce-lv2 ];
+    lv2Path = lib.makeSearchPath "lib/lv2" ls2Packages;
   in lib.mkIf (cfg.setupAsahiSound && cfg.enable) (lib.mkMerge [
     {
       # can't be used by Asahi sound infrastructure
@@ -36,13 +37,13 @@
         pulse.enable = true;
 
         configPackages = [ asahi-audio ];
-        extraLv2Packages = [ lsp-plugins pkgs.bankstown-lv2 ];
+        extraLv2Packages = lv2Packages;
 
         wireplumber = {
           enable = true;
 
           configPackages = [ asahi-audio ];
-          extraLv2Packages = [ lsp-plugins pkgs.bankstown-lv2 ];
+          extraLv2Packages = lv2Packages;
         };
       };
 
--- a/apple-silicon-support/packages/alsa-ucm-conf-asahi/default.nix
+++ b/apple-silicon-support/packages/alsa-ucm-conf-asahi/default.nix
@@ -4,14 +4,14 @@
 }:
 
 (alsa-ucm-conf.overrideAttrs (oldAttrs: let
-  versionAsahi = "5";
+  versionAsahi = "8";
 
   srcAsahi = fetchFromGitHub {
     # tracking: https://src.fedoraproject.org/rpms/alsa-ucm-asahi
     owner = "AsahiLinux";
     repo = "alsa-ucm-conf-asahi";
     rev = "v${versionAsahi}";
-    hash = "sha256-daUNz5oUrPfSMO0Tqq/WbtiLHMOtPeQQlI+juGrhTxw=";
+    hash = "sha256-FPrAzscc1ICSCQSqULaGLqG4UCq8GZU9XLV7TUSBBRM=";
   };
 in {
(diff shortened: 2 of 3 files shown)


kitten:
Interesting! Unfortunately, KDE doesn't seem to present me with input audio mode options at all on my device (J415) and I don't know enough about Wireplumber and Pipewire to debug this, or look into that. I've been using these kinds of changes on my own branch and removing the flags has been working fine for mic support. Today I've pulled in the new Kernel tag which does this and everything seems fine: https://github.com/kitten/nixos-apple-silicon/commit/4789697f6324a14ac14df97a5590f71cc8e47c26

But again, I don't really know enough about this to be able to tell whether all is as intended. All I can say is that audio output and input seem to work fine and sound passable.

kitten:
It's looking to me then like these logs are quite true in your case, right?
> Mar 21 12:37:49 m1 wireplumber[2222]: wp-event-dispatcher: <WpAsyncEventHook:0x8825680> failed: <WpSiStandardLink:0x8d69ee0> link failed: 6 of 6 PipeWire links failed to activate
Because "Built-in Audio" (on my device) refers to the headphone jack, not the device's speakers and microphone

apple-silicon-support/modules/sound/default.nix
```diff
@@ -21,7 +21,8 @@
 
     lsp-plugins-is-safe = (pkgs.lib.versionAtLeast lsp-plugins.version "1.2.14");
 
-    lv2Path = lib.makeSearchPath "lib/lv2" [ lsp-plugins pkgs.bankstown-lv2 ];
+    lv2Packages = with pkgs; [ lsp-plugins bankstown-lv2 triforce-lv2 ];
+    lv2Path = lib.makeSearchPath "lib/lv2" ls2Packages;
```
Lederstrumpf, line 25: typo `lv2Packackes`, but also is unused var.

```suggestion
```

Lederstrumpf:
Thanks for working on this - a headset-less future is finally in sight :D

> I think we can wait the day or two for triforce-lv2 to arrive in the appropriate channels.
> 
> Did you test that this actually works with the 6.13.7-4 kernel? With similar changes, I am getting input on the "Pro Audio" mode (which I think disables the dsp?) but no input on the "Default" mode.

Have the same behavior here - input works on the Pro Audio device, else not.
Tested with `arecord` & vcalls in browsers.

> Unfortunately, KDE doesn't seem to present me with input audio mode options at all on my device (J415) and I don't know enough about Wireplumber and Pipewire to debug this, or look into that.

Interesting - do you only one built-in device listed under `pavucontrol`, and only the default option presented with `pavucontrol->Configuration->Built-in Audio->Profile`?

kitten:
image
image

Had a look with `pavucontrol`. I'm seeing the devices as expected, one input and output for the built-in audio jack, one output for the speakers, one for the microphones.

The only difference I can think of would be how the `asagi-6.13.7-4` kernel is compiled. On my `edge` branch I replaced the Kernel config with one derived from the base config and Nix defaults (it's larger and takes longer to build, but I basically don't want to maintain a full Kernel config in this fork). Not sure if there's other things it could be.
But from the original logs, obviously, this stands out:
> spa.alsa: 'hw:0,2': capture open failed: Device or resource busy

kitten:
I tried to see if this may be related to my other pipewire/wireplumber configs, but disabling them didn't break the sounds devices for me: https://github.com/kitten/system/blob/dd63398c75f573104651790673750a5448ece9b0/modules/desktop/services.nix#L56-L135

I could try to switch my config over to this branch exactly, update the kernel, and basically do a check with the `tpwrules/nixos-apple-silicon` kconfig though

kitten:
This might be a bit of an obvious thing to point out, but have y'all checked the support table on `asahi-audio` against your device ([which now has a microphone section](https://github.com/AsahiLinux/asahi-audio/?tab=readme-ov-file#currently-supported-devices-microphones))?

There's also active changes being made to some unlisted devices on the `mics` branches here:
- https://github.com/AsahiLinux/asahi-audio/commits/mics
- https://github.com/AsahiLinux/alsa-ucm-conf-asahi/commits/mics/

flokli:
`triforce-lv2` has landed in nixpkgs (see https://nixpk.gs/pr-tracker.html?pr=391796 and https://nixpk.gs/pr-tracker.html?pr=391841.

Can you drop the part adding it to the overlay, and bump the pinned nixpkgs past that commit?

flokli:
I sucessfully tested this, alongside https://github.com/tpwrules/nixos-apple-silicon/pull/284, https://github.com/tpwrules/nixos-apple-silicon/pull/289, https://github.com/tpwrules/nixos-apple-silicon/pull/290 and https://github.com/tpwrules/nixos-apple-silicon/pull/292, and got working microphones (after recreating new APFS / EFI partitions, which bumped my `/proc/device-tree/chosen/asahi,os-fw-version` to `13.5`. With the `12.4` previously `aop` wasn't able to probe).

maan2003:
do you get echo from speakers? I have lot of echo. do we need explicit echo canceller

flokli:
> do you get echo from speakers? I have lot of echo. do we need explicit echo canceller

This is out of scope for this repo, we don't have any custom downstream config here. Please discuss it in the upstream Asahi channels. Thanks!

tpwrules:
Thanks very much for the patience. It took a bit until I had the time to rebuild my system with the new firmware but I can confirm this works now!

I merged this manually with some other tweaks and improvements. This should work now (at least on default Plasma 5 with audio setup enabled).

PR #288, opened 2025-03-31
flake: expose linux-asahi-with-rust

flokli:
This makes it easier to build a standalone kernel with rust support enabled, by exposing it as an output.


flokli:
@yuyuyureka mentioned the reasons for which we disable Rust support in our kernel build are outdated (nixpkgs enables rust support as well by now), so we should follow suit. Closing this PR.

PR #289, opened 2025-03-31
flake.lock: bump nixpkgs

flokli:
This brings triforce-lv2, so https://github.com/tpwrules/nixos-apple-silicon/pull/287 can use it.


tpwrules:
Done manually.

PR #290, opened 2025-03-31
linux-asahi: Remove custom rust wrappers

yuyuyureka:
nixpkgs can handle cross-compilation just fine, and in recent nixpkgs versions (< 1 year old) it will never re-build rustc or llvm for cross.


tpwrules:
Thanks for looking into this and for your patience.

The Oxalica Rust overlay is only used by the m1n1 chainloader support as it was an easy way to get the required cross compiler. That has probably rotted a bit (and removing the overlay will certainly fully break it), but please remove that commit from this PR, let's keep this one to focus on the kernel.

PR #311, opened 2025-06-23
asahi-audio: 3.3 -> 3.4; use nixpkgs style

### LINKED ISSUE #306: Audible artifacts on M2 MBA
Running release 2025-05-30, unsure if this repros on older releases. This is the relevant subset of my config:
```nix
{...} : {
  imports =
    [ # Include the results of the hardware scan.
      ./hardware-configuration.nix
      apple-silicon.nixosModules.apple-silicon-support
    ];
  # More stuff here...
  hardware.asahi = {
    enable = true;
    setupAsahiSound = true;
    peripheralFirmwareDirectory = ./firmware;
    useExperimentalGPUDriver = true;
    experimentalGPUInstallMode = "overlay";
  };
  # More stuff here...
}
```
I initially didn't set `setupAsahiSound` to true and heard ocassional audio artifacts. After setting it to true the behavior didn't change.

normalcea:
In addition to updating asahi-audio, this PR improves the derivation builder code for asahi-audio using the current nixpkgs style.

This also brings up the question of formatting which the project does not specify. I used the nixpkgs formatter (`pkgs.nixfmt-rfc-style`) but this project should have a dedicated formatter (via `nix fmt` and `treefmt`). On that issue I'd recommend that this project stick as close to the nixpkgs way of doing things since it's now under `nix-community`. I'd be interested in helping out in that migration.

I haven't tested whether this resolves #306, which may be due to a module issue rather than the package itself.


ghost:
I've created an upstream nixpkgs PR https://github.com/NixOS/nixpkgs/pull/419428 which is probably a better move overall.

flokli:
I'd personally prefer upstreaming this to nixpkgs and then dropping the expression from here entirely. We can also backport the package addition to 25.05 afterwards, if 25.05 support is a concern.

ghost:
https://github.com/NixOS/nixpkgs/commit/da7cdb5c37972f2ff2608d25320e8479bbe3b450 upstreamed! Many thanks. Now the transition will have to be made to remove `asahi-audio` from the overlays.


> if 25.05 support is a concern.

Most (if not all) users of nixos-apple-silicon are on `nixos-unstable` so I don't think that would be necessary.

PR #313, opened 2025-06-25
modules/mesa: remove dead code

flokli:
The isMode helper function is unused, it can be removed.


ghost:
This would be caught by a linter like [deadnix](https://github.com/astro/deadnix) which is used by projects like [disko](https://github.com/nix-community/disko/blob/master/treefmt.toml)

- https://github.com/nix-community/nixos-apple-silicon/pull/314

PR #326, opened 2025-07-23
linux-asahi: asahi-6.14.8-1 -> asahi-6.15.10-3

flokli:
simple-mfd-spmi and nvmem_spmi_mfd are gone, the SPMI nvmem support was merged as apple_nvmem_spmi for v6.16, so we need that module.


ghost:
I have been using these two patches on my own branch for a couple of days now on a MBA 2020, I have not noticed any regressions from typical use yet.

waltmck:
I have also been testing this and it seems to work fine.

tpwrules (changes requested):
* Fedora is still on 6.14.8 AFAICT
* This appears to break the RTC on MacBook Pro M1 Max. There is no `/dev/rtc` anymore and dmesg says `[   14.307802] platform macsmc-rtc: deferred probe pending: macsmc-rtc: Failed to get rtc_offset NVMEM cell`. I noticed cause the time in the installer was December 2024.

Not done any debugging yet, but this is not ready to merge. Does seem to work otherwise though.

tpwrules:
The RTC issue might only show up on the in-tree nixpkgs. If upgrading fixes it (I will file a PR for that soon) my soul doesn't like it but it's probably not worth a lot of debugging effort.

flokli:
I tested on an M2 Air, with nixpkgs c87b95e25065c028d31a94f06a62927d18763fdf, and it does come up:

```
Dez 19 22:27:50 m2air kernel: macsmc-rtkit 23e400000.smc: Failed to create device link (0x180) with supplier 0-0e for /soc/smc@23e400000/rtc
Dez 19 22:27:51 m2air kernel: macsmc-rtc macsmc-rtc: registered as rtc0
Aug 02 19:12:37 m2air kernel: macsmc-rtc macsmc-rtc: setting system clock to 2025-08-02T17:12:37 UTC (1754154757)
```

We can probably wait until Fedora switched too.

tpwrules:
Turns out this isn't a nixpkgs issue, it's just incompatible device trees between 6.14.8 and 6.15.8. There are indeed differences in how the nvmem works.

Testing order and the fact that the NixOS generation system doesn't change them confused me. Now that my system device trees are from 6.15.8, my 6.14.8 generations no longer have an RTC.

This shouldn't affect normal upgrades. The guide already has instructions to set the time from the network so this is probably fine, just annoying.

dramforever:
And even asahi-6.16.0-2 is out. Seems to work for me.

We should probably still follow fedora... but at the same time this PR is nice for anyone wanting to try out the latest but not necessarily greatest.

waltmck:
I just asked in the Fedora Asahi matrix, and according to some Fedora maintainers [this BTRFS fix](https://lore.kernel.org/linux-btrfs/20250730201752.GA909565@zen.localdomain/T/) was blocking the upgrade to 6.15. That patch just made it to stable in Fedora today, so the kernel in Fedora should be upgraded soon.

An aside: if we are willing to tolerate months-long kernel update delays to support BTRFS, why not tolerate weeks-long kernel update delays to support ZFS?

dramforever:
> An aside: if we are willing to tolerate months-long kernel update delays to support BTRFS, why not tolerate weeks-long kernel update delays to support ZFS?

Mostly just that the Asahi Linux team works most closely with shipping Fedora, so that's what we're following

tpwrules:
Looks like we will need this for current master due to Rust changes. Please rebase with the `asahi-6.16.1-1` tag. That also looks like it has the btrfs fix on top.

Would also appreciate squashing into one commit.

tpwrules:
Updated to the latest Fedora release (as of like an hour ago) which includes the btrfs fix and will build on current nixpkgs master. Tested on my system that everything seems to work and the installer still boots. Would welcome review.

reinthal:
Will perform reinstall of nixos on apple m1 beginning of next week. Looking foward to this update.

PR #328, opened 2025-08-02
docs: add contributing.md

flokli:


ghost:
**suggestion**: alternative release model proposal

I think we should retire the `release-YYYY-MM-DD` tag pattern and instead mimic the nixpkgs pattern. Instead of having `release-YYYY-MM-DD` we can just have merges to `nixos-unstable` that acts as a rolling release, then to releases that follow nixos-stable

``` text
contributions/PRs -> main -> nixos-unstable  ->  nixos-{YY-MM}
```

I think there's a strong case to be had that Asahi users don't actually want to be withheld features from newer kernel. At this point, this repository holds downstream `hardware.asahi` modules, vendored uboot/kernel and the bootstrap installer and guide.

We can instead of a evolving changelog for the next release in terms of changes similar to how NixOS releases are done (where release notes for the next major release are added)

**question**: clarification for "release-YY-MM" branches

On these branches, will the nixpkgs input be changed to match releases (example: `"github:nixos/nixpkgs/nixos-25.11"`) or is it going to be frozen at a specific commit in unstable?

**suggestion**: testing instructions

> - Each installer is built, booted from it, and a re-installation is performed,
   to detect problems with booting, networking, and stuff essential to the new
   user experience. After that, important features are tested.


I think the model of machine should also be specified, whether it's an M1 or later, or a Mac Mini vs. the Pro or Air. Also how many models were tested.

We should also specify which desktop environment (or just sole compositor) will be used in testing, we currently still have a snippet in `uefi-standalone.md` that uses XFCE (X11) as an example DE even though Asahi upstream recommends against using X11.

**nitpick**: wording on testing instructions

"stuff essential to the new user experience" seems to be related to the important features bit after it, which is confusing.

**question**: clarify git log mention

> The guide is checked to be up to date (check git log for example)

I don't fully understand what this means.

**question**: what does "good" mean?

> Once everything is good,

What is the threshold for considering a release "good?" Is it when the nixos-apple-silicon maintainers give their thumbs up or does there need to be a form of community consensus from other members?

tpwrules:
@flokli @normalcea updated and most of the feedback has been addressed

I strongly disagree with removing tagged releases. I think they are essential to providing a validated starting point, especially as development becomes more asynchronous. They are also good for the installer build process. But I think most users do not have to track them directly.

PR #329, opened 2025-08-05
flake: overhaul flake

ghost:
### Outputs pattern

I noticed that the flake used `self` as a self-referential attribute but the more idiomatic way to use it is with `inherit (self) outputs` which allows for dropping `rec` and makes self-referring much easier.

### Dropping treefmt-nix

In my research while looking at the CI for nixpkgs, I discovered the `nixfmt-tree` package which fills in all the messy details for us while also allowing for dropping the treefmt-nix input.

For more information see: <https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/by-name/ni/nixfmt-tree/package.nix#L60>

For the PR linting workflow: we can use `nix flake check` as always.

### devShells/Direnv

In addition, I added direnv support and devshells which contains only `treefmt` in its environment to assist in contributions.

### Default package attribute:

The next thing that the devshell also pointed out is that there wasn't a default package attribute for this flake, so I used the installer-boostrap for it. This allows for one-liners like this to be used:

``` shellsession
$ nix build github:normalcea/nixos-apple-silicon/flake-rework
evaluating derivation 'github:normalcea/nixos-apple-silicon/flake-rework#packages.aarch64-linux.default'
[2/0/118 built, 0.0 MiB DL] building systemd-257.5 (patchPhase):
```

So we can simplify the cloning step for flakes users (which afaik is in the majority) and also have a one liner for building the ISO which can shorten the install documentation which has users build each package separately. 

### flatten `nixpkgs.url`

Just for consistency, I don't see any need for any additional flake inputs besides nixpkgs and flake compat.

## Testing

For testing this PR, I ran `nix flake check` as well as `nix flake show` and `nix flake metadata`

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -2,10 +2,7 @@
   description = "Apple Silicon support for NixOS";
 
   inputs = {
-    nixpkgs = {
-      url = "github:nixos/nixpkgs/nixos-unstable";
-    };
-
+    nixpkgs.url = "github:nixos/nixpkgs/nixos-unstable";
     flake-compat.url = "github:nix-community/flake-compat";
   };
 


tpwrules:
> Outputs pattern

This seems reasonable. Do you have info on the idiom? Does this increase evaluation time?

> Dropping treefmt-nix

This also seems good, I like less inputs. What does this do to `nix fmt`? I assume this doesn't change which `nixfmt` executable is used.

> flatten nixpkgs.url

Reasonable.

> devShells/Direnv
> Default package attribute:

I don't understand the advantages here, assuming `nix fmt` still works. A one-liner works today with `nix build github:nix-community/nixos-apple-silicon#installer-bootstrap`.

The docs were written long ago. Even back then though I think the majority of users should use the pre-built ISOs though that may change as we change the release model. The docs certainly need updating, I don't think that thought has ever been particularly clear.

ghost:
> This seems reasonable. Do you have info on the idiom? Does this increase evaluation time?

- https://discourse.nixos.org/t/who-is-self-in-flake-outputs/31859/4

`self` is referring to the whole flake as a whole, since we only want to refer to flake outputs, it's more idiomatic to do it this way. 

```shellsession
$ nix eval .#self.
.#self.checks            .#self.inputs            .#self.outPath           .#self.sourceInfo
.#self.devShells         .#self.lastModified      .#self.outputs           .#self.submodules
.#self.dirtyRev          .#self.lastModifiedDate  .#self.overlays          .#self._type
.#self.dirtyShortRev     .#self.narHash           .#self.packages          
.#self.formatter         .#self.nixosModules      .#self.self              
```

> This also seems good, I like less inputs. What does this do to nix fmt? I assume this doesn't change which nixfmt executable is used.

It doesn't, nixfmt-tree is simply just a nixpkgs way of creating a treefmt.toml for use with `treefmt` running `nix fmt` has no changes.

> I don't understand the advantages here

Many users (including me) use direnv to automatically enter a nix shell environment within their editors. nix-shell complains if there isn't a default package set and so setting the installer iso as the default package resolves that.

flokli:
> Many users (including me) use direnv to automatically enter a nix shell environment within their editors. nix-shell complains if there isn't a default package set and so setting the installer iso as the default package resolves that.

I'm not sure the installer iso is a reasonable default package setting. nixos-apple-silicon provides a bunch of individual packages, but its main output is mostly the module code, which is consumed from other NixOS code. Maybe it's fine to leave that warning for now. Can you rebase this PR (it needs rebasing) and drop 50b90e160c030de55bb9e3534d515a5b089dfd41?

tpwrules:
Sorry, I thought I commented, but my concern with making the installer as the dev shell is that the user would have to build all the dependencies (kernel, etc.) just to start the environment.

ghost:
Dropped the default package commit and rebased on current main.

ghost:
Force pushed and rebased on current main

montchr (approved):

PR #303, opened 2025-05-26
remove mesa-asahi-edge

flokli:
Since mesa 25.1, support for asahi is enabled by default. This means,
the nixpkgs mesa is sufficient and we don't need any custom mesa, or
methods to replace it.

In addition to the now ineffective
hardware.asahi.useExperimentalGPUDriver option, this also removes the
hardware.asahi.experimentalGPUInstallMode option, which was already
deprecated before.

Fixes #338.

PR #331, opened 2025-08-10
release 2025-08-10

tpwrules:
First one I'm not just pushing tags for!

Tested on 2020 Macbook Pro 15" M1 Max that it builds, boots, and installs. Using a Plasma 6 desktop environment. Graphics and webcam and sound all work. I haven't identified any issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ Please see the documentation and guide below to get started.
 ## Documentation
 
 * [Release Notes](docs/release-notes.md)
-* [Setup, Installation, and Maintenance Guide (2025-05-30)](docs/uefi-standalone.md)
+* [Setup, Installation, and Maintenance Guide (2025-08-10)](docs/uefi-standalone.md)
 
 ## Credits
 
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -2,6 +2,16 @@
 
 This file contains important information for each release.
 
+## 2025-08-10
+
+This release updates nixpkgs.
+
+Rust support is now the default in the kernel (even in the installer). This
+release also fixes a build issue with asahi-fwextract.
+
+As of this release, the release model is now documented [here](https://github.com/nix-community/nixos-apple-silicon/blob/main/docs/release-process.md). Development
+is now shifting directly to the main branch instead of discrete tagged releases.
+
 ## 2025-05-30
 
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -9,7 +9,7 @@ Please see the documentation and guide below to get started.
 ## Documentation
```
flokli, line 9: Do we want to explicitly point out that stable users should pick the latest release-mm-yy branch here, to make it more visible? It might be a bit too hidden if only on the release process documentation, which is something users might  rarely look at.

docs/release-notes.md
```diff
@@ -2,6 +2,16 @@
...
 
+## 2025-08-10
+
+This release updates nixpkgs.
+
+Rust support is now the default in the kernel (even in the installer). This
+release also fixes a build issue with asahi-fwextract.
+
+As of this release, the release model is now documented [here](https://github.com/nix-community/nixos-apple-silicon/blob/main/docs/release-process.md). Development
+is now shifting directly to the main branch instead of discrete tagged releases.
```
flokli, line 13: And mention that the main brauch is only tested to work with unstable.

README.md
tpwrules: I'm not sure that's necessary, there is already a release link in the guide that will show the latest first. And now users can use the main branch instead of a tagged release.

I don't think anyone's had a problem with it before.

flokli: Ok

docs/release-notes.md
flokli: Ok

flokli (approved):

tpwrules:
Okay, it turns out I screwed this up and didn't test the proper nixpkgs commit so this left us with #338 . Oops...

This is because I don't use the flake so I have to manually change my system's nixpkgs and I guess I messed it up or forgot. Unsure why flokli didn't manage to turn that up either.

flokli:
Yeah, shit happens, sorry. I also only tested with my nixpkgs revision, as I don't use the flakes interface. With the nix-community Buildbot integration and their binary cache setup, I'd really like to start adding more of this to CI (at least the kernel), so it becomes much easier to test if the substituter is configured.

Let's get https://github.com/nix-community/nixos-apple-silicon/pull/303 landed, cut another release with it and then start to work on CI bits to make this less of a hazard :-)

PR #333, opened 2025-08-11
packages: use nixpkgs upstream counterparts to asahi packages

normalcea:
This PR removes `alsa-ucm-conf-asahi` `asahi-audio` and `asahi-fwextract` in favor of the upstream nixpkgs derivations. 

m1n1 is left alone because the nixpkgs upstream version does not have full parity and would require changes to module configurations. I want m1n1 in nixpkgs to be a suitable replacement so reviews on https://github.com/NixOS/nixpkgs/pull/430738 would contribute to this.

Mesa requires a nixpkgs upgrade and can be handled here: https://github.com/nix-community/nixos-apple-silicon/pull/303


ghost:
I have updated this PR to include updating the nixpkgs input to include m1n1-1.5.0 and added m1n1 from nixpkgs. This removes the the outputs from `withTools` and chainloading but those aren't needed by the vast majority of users.

tpwrules:
Please move the m1n1 changes to a separate PR for discussion.

ghost:
@tpwrules 
 > Please move the m1n1 changes to a separate PR for discussion.

m1n1 changes submitted here (#349) and removed from this PR

tpwrules:
Looks like these match pretty well. There is an inconsequential asahi-fwextract upgrade and a slight rearrangement of the asahi-audio package but they seem to work properly still.

PR #335, opened 2025-08-12
linux-asahi: update config

normalcea:
This PR updates the kernel config to match the 6.14.8 config https://github.com/asahi-alarm/PKGBUILDs/blob/e76f14aa9dce2bdff94edb44956bf75f7d3c7276/linux-asahi/config and the fedora-asahi-6.14 branch (https://gitlab.com/fedora-asahi/kernel-asahi/-/tree/fedora-asahi-6.14)

In addition, it cleans up unused codepaths within `linux-asahi` 

I was able to use this kernel config for my MBA 2020 and I haven't noticed any obvious regressions.


flokli:
Somewhat conflicts with https://github.com/nix-community/nixos-apple-silicon/pull/326, but it's ok if we just delay that one until Archlinuxarm / Fedora follow.

tpwrules:
Is this going to substantially increase the build time? Rather than syncing with Fedora we should probably just use the default NixOS kernel build if we are going for something “full-fat”. But this might be okay now once we get the community cache working.

ghost:
+1 on full kernels, it is a negative that users have to patch their kernel to get things like bcachefs working.

waltmck:
> +1 on full kernels, it is a negative that users have to patch their kernel to get things like bcachefs working.

I agree with this. Additionally, the switch to using a full kernel would be a good opportunity to revisit the current policy of closely following Fedora. For example, the 6.15 Asahi kernel has been ready for more than 3 months now (and 6.16 has been ready for a couple of weeks), but Fedora has been stuck on 6.14 because they were blocked on packaging a BTRFS fix.

I don't see a compelling reason why this Fedora-specific blocker should prevent NixOS from upgrading their kernel version. Perhaps going forward we should evaluate these things on a case-by-case basis.

Alternatively, if we want to follow some sort of upstream guidelines, a better target might be Gentoo's [asahi-overlay](https://github.com/chadmed/asahi-overlay) which is maintained by two Asahi project board members. That repo currently appears to support 6.14.8, 6.15.7, and 6.16.0 kernels.

ghost:
Multiple kernels sounds like a decent compromise, it would need accompanying documentation as well to inform users the differences between each one.

dramforever:
`asahi-6.16.1-1` has cherry-picked the btrfs fix, and upstream 6.16.2 should just have it

It's not Fedora specific. This affects all BTRFS users.

yuyuyureka:
Consider #341 an alternative, someone could do benchmarking of how long the build of the old config vs Fedora based vs NixOS based takes.

ghost:
I updated the config with the extra configuration options added in this https://github.com/asahi-alarm/PKGBUILDs/commit/c3b696ca7c6af53c8a90f4964b9a3613508d530b commit which adds the PS4 HID driver.

ghost:
Resolved merge conflicts and rebased on main.

flokli:
@normalcea would it make sense to close this in favor of https://github.com/nix-community/nixos-apple-silicon/pull/341 ?

dramforever:
#341 would cause pain for users of lower-powered apple silicon devices, unless we can get a trustworthy binary cache solution going

yuyuyureka:
Superseeded by #341

Issue #294, opened 2025-04-30
unable to launch Hyprland or Niri session since release-2025-04-27-2


astratagem:
Since updating to the latest release `release-2025-04-27-2`, I have not been able to successfully launch ~a Sway or~ Hyprland session, which I do from `regreet` using `uwsm`.

I have the output of `journalctl --boot=-1 --all` available here:

https://gist.github.com/montchr/9305071447a80040c3766c81f5a036a8

I am able to launch a Sway (UWSM) session ~for a few seconds, with cursor visible and responsive, but otherwise the session seems non-functional.  Eventually, the session ends and kicks me back to regreet~ [edit: full stop -- Sway is fine].  Then, I try to launch Hyprland (UWSM), but after the session starts, all I see is a black screen.  I am not able to switch TTY after that point, and so I can only force shutdown by holding the power button.

The logs show a few notable messages, though I am not sure what is actually relevant.  The greeter user session for greetd starts at [L1742](https://gist.github.com/montchr/9305071447a80040c3766c81f5a036a8#file-journal-log-L1742).  [Edit: Ignore the part of the log where Sway fails due to a missing config, which was a red herring]  Strangely, it looks like Hyprland thinks it has started properly, but by the time that happens, UWSM has given up and I am stuck in the void.

FWIW I posted a comment about this in https://github.com/tpwrules/nixos-apple-silicon/pull/284#issuecomment-2816860786 but I get the sense after hours of troubleshooting and searching for answers that this is probably an edge case 😅 

Also worth noting that I have encountered the same behavior without using UWSM.  

Update: Sway is actually fine -- it failed to launch because of a missing config, which is indicated in the log -- but Hyprland still fails.

jmastr:
Cannot confirm:
```
$ neofetch
          ▗▄▄▄       ▗▄▄▄▄    ▄▄▄▖            julian@apple-macbook-pro
          ▜███▙       ▜███▙  ▟███▛            ------------------------
           ▜███▙       ▜███▙▟███▛             OS: NixOS 25.05pre793735.979daf34c8ca (Warbler) aarch64
            ▜███▙       ▜██████▛              Host: Apple MacBook Pro (14-inch, M1 Pro, 2021)
     ▟█████████████████▙ ▜████▛     ▟▙        Kernel: 6.13.5-asahi
    ▟███████████████████▙ ▜███▙    ▟██▙       Uptime: 2 mins
[...]

PR #341, opened 2025-08-23
de-vendorify kernel config (build full NixOS kernel)

yuyuyureka:
I've been running this (NixOS-derived) kernel config for many months.

The downside is it takes a few times longer to build, since it compiles all possible modules. But this should no longer be a problem now that we have a CI and binary cache.

### DIFF
--- a/apple-silicon-support/packages/linux-asahi/default.nix
+++ b/apple-silicon-support/packages/linux-asahi/default.nix
@@ -17,15 +17,15 @@ let
     buildLinux rec {
       inherit stdenv lib;
 
-      version = "6.15.10-asahi";
+      version = "6.16.5-asahi";
       modDirVersion = version;
-      extraMeta.branch = "6.15";
+      extraMeta.branch = "6.16";
 
       src = fetchFromGitHub {
         owner = "AsahiLinux";
         repo = "linux";
-        tag = "asahi-6.15.10-3";
-        hash = "sha256-au/v0bLzBaAMscfk47MZpyiG8pomw08qlT1RjVO9/x4=";
+        tag = "asahi-6.16.5-2";
+        hash = "sha256-W30x9aoU5ltkda/flfhdGbMHVYCpNaneYvhbh7HiG+c=";
       };
 
       kernelPatches = [


dramforever:
If we drop that all the way down to these (which are necessary to get `DRM_ASAHI=m` it seems?)

```
    # For DRM_ASAHI
    DRM = yes;
    ARM64_16K_PAGES = yes;
```

the difference with yours is (`simple -> yours`)

```
CONFIG_HID_APPLE                y -> m
CONFIG_APPLE_WATCHDOG           m -> y
CONFIG_DRM_VGEM                 m -> n
CONFIG_APPLE_M1_CPU_PMU         n -> y
```

My take on these, from the perspective of whether the "simple" config needs changing:

- `HID_APPLE=y` is fine.
- I don't know which of `y` or `m` for `APPLE_WATCHDOG` is better. I *think* what happens if we get to Linux and it doesn't initialize `apple_wdt` in time, the machine forcibly reboot, and after three strikes it goes to recovery. So if we do `m` here, I think we should also add it to `boot.initrd.availableKernelModules` so you have like more than a minute or something to do initrd debugging.
- `DRM_VGEM=m` should be fine. It helps with software renderers that need it like llvmpipe, and doesn't hurt otherwise.
- `APPLE_M1_CPU_PMU=n` is annoying. it can't be `m` because it uses non-`EXPORT_SYMBOL` symbols. I personally would have it at `y` as long as we build our own kernel. In the far future where upstream kernel works on M1 I think we wouldn't need n-a-s by that point.

yuyuyureka:
@dramforever what is this "simple" config you are talking about, where can I find it?

Can you perhaps make a specific suggestion using the diff feature in github?

yuyuyureka:
Ah, you dropped all options from the extraStructuredConfig except those two and then diffed the generated config, meaning all other statements are no-ops?

I would think we also need `ARCH_APPLE = yes;`?

I also don't understand how HID_APPLE can be "y" when not specified, since it can be built as a module the generate-config.pl would always opt for that.

dramforever:
`HID_APPLE` is `default !EXPERT || SPI_HID_APPLE`, which is `default y`, which the config script thing honors instead of making that `m`

dramforever:
Oh, missed this: I actually don't know why `ARCH_APPLE` defaults to `y`, but it does, and in fact does for all `ARCH_*`

yuyuyureka:
Removed redundant config (and some more; I think DRM_SCHED-related things are also differing, but it seems to work fine with the simple version on my machine).

Also attempted integrating it into the CI.

tpwrules:
Looks like the CI build is working, thanks for the experiment. I will test it more thoroughly later. We should definitely not be building for x86-64 though, although I guess that's neat that it worked.

yuyuyureka:
From what I can tell we're not building _for_ x86_64, but just _on_ x86_64, which is good since it will cache the derivation for cross-compilation users.

waltmck:
I just built this with the latest nixpkgs and have a few issues:

1. There is no GPU acceleration (Vulkan nor OpenGL). I saw that there was a rust-related issue and the commit fixing it is not in this branch, so maybe this is to be expected.
2. `options hid_apple swap_fn_leftctrl=1` no longer works to swap the control and function keys.

dramforever:
> I just built this with the latest nixpkgs and have a few issues:
> 
> 1. There is no GPU acceleration (Vulkan nor OpenGL). I saw that there was a rust-related issue and the commit fixing it is not in this branch, so maybe this is to be expected.

Yes, this is a known issue worked around in #343 and should be fixed for properly in https://github.com/NixOS/nixpkgs/pull/436245

> 2. `options hid_apple swap_fn_leftctrl=1` no longer works to swap the control and function keys.

This is expected for `HID_APPLE=y`. You'd need to move/add  that to `boot.kernelParams` like:

```nix
boot.kernelParams = [ "hid_apple.swap_fn_leftctrl=1" ];
```

yuyuyureka:
Ohh I guess that's a reason to keep HID_APPLE=m...

I just rebased, so the Rust workaround should be included now

waltmck:
We talked about this in the matrix chat, but I will summarize my results testing the latest version of this PR here for transparency:

- The built-in speakers are very quiet compared to the existing kernel. This same problem happened with @zzywysm's custom kernel, and the fix was to [tweak some config options](https://github.com/nix-community/nixos-apple-silicon/commit/93a4cc1e9cd4bbf97fe4c2a70cac35dc05d5ae8e). I do not have insight into why those changes were necessary, and I have not yet tested whether they fix the issue on 6.16 as well.

- IWD 802.1x support is broken---when I try to connect to such a network, I get `Error loading client private key`. The fix is to add
  ```nix
  boot.kernelModules = ["pkcs8_key_parser"];
  ```
  This may be an issue with the IWD module, but I do not have any non-ARM wifi-capable hardware to test this on. I will file an upstream bug when I have the time.

- The kernel no longer builds with `-march=armv8.6a`, which is a shame although it probably isn't supported. The error message is
  ```
  /build/cc2qPCGH.s: Assembler messages:
  /build/cc2qPCGH.s:23760: Error: selected processor does not support `retaa'
  /build/cc2qPCGH.s:24801: Error: selected processor does not support `retaa'
  /build/cc2qPCGH.s:25196: Error: selected processor does not support `retaa'
  /build/cc2qPCGH.s:25487: Error: selected processor does not support `retaa'
  /build/cc2qPCGH.s:26704: Error: selected processor does not support `retaa'
  /build/cc2qPCGH.s:27997: Error: selected processor does not support `retaa'
  /build/cc2qPCGH.s:28784: Error: selected processor does not support `retaa'
  /build/cc2qPCGH.s:29808: Error: selected processor does not support `retaa'
  /build/cc2qPCGH.s:29923: Error: selected processor does not support `retaa'
  /build/cc2qPCGH.s:30010: Error: selected processor does not support `retaa'
  /build/cc2qPCGH.s:30034: Error: selected processor does not support `retaa'
  /build/cc2qPCGH.s:30232: Error: selected processor does not support `retaa'
  /build/cc2qPCGH.s:30423: Error: selected processor does not support `retaa'
  /build/cc2qPCGH.s:30510: Error: selected processor does not support `retaa'
  /build/cc2qPCGH.s:30721: Error: selected processor does not support `retaa'
  /build/cc2qPCGH.s:31037: Error: selected processor does not support `retaa'
  /build/cc2qPCGH.s:32079: Error: selected processor does not support `retaa'
  /build/cc2qPCGH.s:32271: Error: selected processor does not support `retaa'
  /build/cc2qPCGH.s:32528: Error: selected processor does not support `retaa'
  /build/cc2qPCGH.s:32768: Error: selected processor does not support `retaa'
  /build/cc2qPCGH.s:33037: Error: selected processor does not support `retaa'
  /build/cc2qPCGH.s:33298: Error: selected processor does not support `retaa'
  /build/cc2qPCGH.s:33508: Error: selected processor does not support `retaa'
  /build/cc2qPCGH.s:33710: Error: selected processor does not support `retaa'
  /build/cc2qPCGH.s:33958: Error: selected processor does not support `retaa'
  /build/cc2qPCGH.s:34176: Error: selected processor does not support `retaa'
  /build/cc2qPCGH.s:34630: Error: selected processor does not support `retaa'
  /build/cc2qPCGH.s:34872: Error: selected processor does not support `retaa'
  /build/cc2qPCGH.s:35147: Error: selected processor does not support `retaa'
  /build/cc2qPCGH.s:35415: Error: selected processor does not support `retaa'
  make[8]: *** [../scripts/Makefile.build:287: drivers/net/ethernet/marvell/octeontx2/nic/otx2_pf.o] Error 1
  ```

waltmck:
i fixed the audio issue by building some things as modules, see [my branch](https://github.com/waltmck/nixos-asahi/tree/minimize-patches).

yuyuyureka:
I think what's still missing here is a somewhat trustworthy binary cache & CI. At the moment, I am not convinced I would want to build even the installer images in the nix-community CI which builds derivations from untrusted PRs.

yuyuyureka:
> i fixed the audio issue by building some things as modules, see [my branch](https://github.com/waltmck/nixos-asahi/tree/minimize-patches).

Thanks Walter! I squashed your commits together and added them to my branch.

ghost:
Any update on this? I would really like to see this merged.

waltmck:
This kernel takes 2-3 hours to compile on my M2 air, compared to around 20 minutes for the minimal one, so I think that a community cache is necessary before it can be merged. Otherwise, the kernel works fine (I have been using it).

ghost:
For me it was less than 2 hours (1 hour 49 minutes) on my MBA m1 2020. There's probably a lot of factors that affect how long the build process takes. 

A step towards a binary cache would be to consolidate n-a-s to what we only need: See https://github.com/nix-community/nixos-apple-silicon/pull/349 and https://github.com/nix-community/nixos-apple-silicon/pull/333 and https://github.com/nix-community/nixos-apple-silicon/pull/329/

waltmck:
> For me it was less than 2 hours (1 hour 49 minutes) on my MBA m1 2020. There's probably a lot of factors that affect how long the build process takes.
> 
> A step towards a binary cache would be to consolidate n-a-s to what we only need: See #349 and #333 and #329

Upstreaming m1n1 and other components to nixpkgs would be nice and is a worthy effort, but they barely add any compile time so I don't think that they should be considered a blocker for moving the vender-specific kernel build to a binary cache. Maybe someone involved in the nix-community infra could tell us if there is anything blocking that.

yuyuyureka:
The last status here was that we (the nixos-apple-silicon maintainers, but mostly me) were not ok with the idea of encouraging people to use the kernels and install images built on the same CI which also builds arbitrary untrusted derivations (PRs on other nix-community projects).
So either the security concept of the nix-community CI & binary cache needs to be clarified and possibly adjusted to separate untrusted and trusted builds on different builders & signing keys, _or_ we build the kernels and installer images using GitHub Actions and push them to our own cache (hosting for that would need to be figured out).

astratagem:
I've just checked this out in my personal fork and I'm happy to report that it completely resolves whatever was causing #294!  So I would hope that once this is merged, there might be fewer reports of issues with hardware-accelerated Wayland compositors like Niri and Hyprland... what exactly the "fix" was, I have no idea.

ghost:
Some nits, remove unused arguments and also this change means we aren't really tracking the asahi-wip tree or basing off of fedora releases.

``` diff
diff --git a/apple-silicon-support/packages/linux-asahi/default.nix b/apple-silicon-support/packages/linux-asahi/default.nix
index 397edf8..a0a5552 100644
--- a/apple-silicon-support/packages/linux-asahi/default.nix
+++ b/apple-silicon-support/packages/linux-asahi/default.nix
@@ -11,10 +11,9 @@ let
       stdenv,
       lib,
       fetchFromGitHub,
-      fetchpatch,
       buildLinux,
       ...
-    }@args:
+    }:
     buildLinux rec {
       inherit stdenv lib;
 
@@ -23,7 +22,6 @@ let
       extraMeta.branch = "6.16";
 
       src = fetchFromGitHub {
-        # tracking: https://github.com/AsahiLinux/linux/tree/asahi-wip (w/ fedora verification)
         owner = "AsahiLinux";
         repo = "linux";
         tag = "asahi-6.16.5-2";
@@ -65,11 +63,6 @@ let
       ++ _kernelPatches;
     };
 
-  linux-asahi = (callPackage linux-asahi-pkg { }).overrideAttrs (_: {
-    # FIXME: Remove when https://github.com/NixOS/nixpkgs/pull/436245 lands
-    preConfigure = ''
-      export RUST_LIB_SRC KRUSTFLAGS
-    '';
-  });
+  linux-asahi = (callPackage linux-asahi-pkg { });
 in
 lib.recurseIntoAttrs (linuxPackagesFor linux-asahi)

```

The last change requires a flake update since it was merged on Aug 23rd and the nixpkgs in n-a-s is from Aug 19th.

PR #343, opened 2025-08-23
linux-asahi: Fix Rust support

dramforever:
Add a hack to fix Rust support broken by NixOS/nixpkgs#423933

See NixOS/nixpkgs#436245 for Nixpkgs upstream fix


dramforever:
This should fix the GPU driver, whether or not using `experimentalGPUDriver` and whatever kernel config is being used. Please test :)

PR #344, opened 2025-08-23
release 2025-08-23

tpwrules:
A bit embarrassed that the last release was broken. This fixes it. Also updates nixpkgs to lock in some fixes there.

Tested on 2020 Macbook Pro 15" M1 Max that it builds, boots, and installs. Using a Plasma 6 desktop environment. Functioning graphics and sound. I haven't identified any issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ Please see the documentation and guide below to get started.
 ## Documentation
 
 * [Release Notes](docs/release-notes.md)
-* [Setup, Installation, and Maintenance Guide (2025-08-10)](docs/uefi-standalone.md)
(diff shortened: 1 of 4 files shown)


GarrettGR:
I came across this as I was struggling with getting another new machine online so I decided to test it on all of my machines to do some experiments.

I have tested this on both a 13" and 15" M2 Macbook Air and a 14" M2 Max Macbook Pro with both KDE Plasma 6 and Hyprland -- it builds, boots, and installs -- both sound and graphics works as well and I haven't found any issues (although one odd behavior on my Macbook Pro where not all wifi networks were being identified, but I think that may be an issue with my machine).

PR #349, opened 2025-08-31
m1n1: use nixpkgs upstream

ghost:
Replaces the n-a-s version of m1n1 with the one in nixpkgs. The motivation for this change is to set an agreed upon collective interface to m1n1 in nixpkgs rather than having to maintain it here ([link to m1n1 package.nix](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/by-name/m1/m1n1/package.nix))

The m1n1 in nixpkgs uses a custom nix logo instead of the asahi one. This can be overridden via the `withBranding` boolean argument and an alternative custom logo can be obtained by overriding the `postPatch` or other phases to get the chosen files in the correct place. 

Notably besides minor build adjustments, m1n1 in nixpkgs still doesn't have an equivalent to `withTools` and `withChainloading` so that would be a incompatible change for n-a-s.


yuyuyureka:
Will merge after #364

PR #353, opened 2025-09-10
Expose installer as a module

n8henrie:
It would be nice to expose the iso-configuration to flake outputs, which will make it significantly easier for end users to create their own customized installer images.

For example, in my case I have been greatly enjoying Asahi NixOS on ZFS root, and so having ZFS support in the installer makes things much easier!
With this PR, users like myself can easily create their own installer image with ZFS support (and git, for flakes and rsync), like so:

```nix
packages.x86_64-linux.zfs-installer =
    (nixpkgs.lib.nixosSystem {
      inherit system;
      pkgs = import nixpkgs {
        crossSystem.system = "aarch64-linux";
        localSystem.system = system;
        overlays = [ nixos-silicon.overlays.default ];
      };
      modules = [
        nixos-silicon.outputs.nixosModules.apple-silicon-installer
        { hardware.asahi.pkgsSystem = system; }
        (
          { pkgs, ... }:
          {
            boot.supportedFilesystems.zfs = true;
            networking.hostId = "AAAAAAAA";
            environment.systemPackages = with pkgs; [
              git
              rsync
            ];
          }
        )
      ];
    }).config.system.build.isoImage;
```

There may be a way to use some overrides with the existing `installer-bootstrap` output, but if so, I haven't yet figured it out: https://discourse.nixos.org/t/custom-asahi-installer


mio-19:
This is very helpful!


It has merge conflicts

n8henrie:
Merge conflict was super minor, resolved.

flokli:
Not meant to be blocking, but can you create a followup PR describing how to build your own installers? That'd be particularly useful if someone for example wants to enable ZFS (c.f. #363).

Issue #352, opened 2025-09-06
[BUG] Speakers volume always reset to max (100%) after reboot


sezaru:
I noticed that every time I reboot my machine (I think it happens if I just logout and login again), my speakers volume will reset to 100%.

After troubleshooting a little bit, I found that the issue seems to be related to the config file that is added to `/etc/pipewire/pipewire.conf.d` `99-asahi.conf`.

Just running `./asahi-diagnose` will complain about that file.

So, what I did to fix it for me was to do the following changes:

First, I will add the setup sound option as per usual:

```nix
hardware.asahi.setupAsahiSound = true;
```

That, by default, will add the above file and trigger the issue, so now I will force it to not do that:

```nix
services.pipewire.configPackages = lib.mkForce [];
services.pipewire.wireplumber.configPackages = lib.mkForce [];

environment.systemPackages = [pkgs.asahi-audio];
```

The first two lines remove `pkgs.asahi-audio` from the `configPackages` for `pipewire` and `wireplumber`, doing that means that the `99-asahi.conf` file will not be copied to `/etc/pipewire/pipewire.conf.d`.

This also means that now there is not link to `pkgs.asahi-audio`, so we manually add it to our `systemPackages`.

After doing that, now the audio will correctly retain its volume after reboots without an issue.

Also, after that change, `asahi-diagnose` will stop complaining that the audio configuration is broken.

So, is there a reason that we are adding that config file? AFAIK that was necessary when they started working on the sound support, but now is not needed anymore and they also recommend that people remove that file manually (with Fedora/Arch Asahi) if they have it to avoid having sound issues.

ghost:
See https://github.com/nix-community/nixos-apple-silicon/pull/357 

This change works, though placing `pkgs.asahi-audio` in `environment.systemPackages` isn't required since the lv2 plugins are only passed into pipewire if they are in the `configPackages` attribute's list. I've proposed changing the sound module to just add the lv2 plugins among other changes.

sezaru:
I want to test that out, but just to make sure I understood it correctly, using that I still can have `services.pipewire.enable = true;` on my end to use pipewire correct?

Edit:

So, I just tried it by copying the changes to my module, basically I have this in my file:

```nix
{
  config,
  lib,
  pkgs,
  ...
}: {
  # From Asahi audio module:

  # can't be used by Asahi sound infrastructure
  services.pulseaudio.enable = lib.mkForce false;
[...]

PR #357, opened 2025-09-19
sound: rework module

normalcea:
Removes the `asahi-audio` configuration options alongside enabling pipewire. This makes `setupAsahiSound` only set up the DSP components and options required rather than enabling sound modules itself.

Should resolve: https://github.com/nix-community/nixos-apple-silicon/issues/352 which explains that the previous implementation is not required anymore. I've tested this for a couple of days on my MBA 2020 and the sound issues with the volume being set to 100% and the GNOME microphone indicator always being present are now gone.

Haven't noticed any notable regressions but I will update this PR if I find them. Also related, we might not need the `pkgs.asahi-audio` package at all if this is all we need and we can retire it from nixpkgs.


ghost:
I'm also not sure if `services.pipewire.wireplumber.extraLv2Packages` should also be set? It seems like it should based on the fact that `asahi-audio` was applied to `configPackages` for both wireplumber and pipewire.

dramforever (changes requested):
This can't be right...

This just removes `pkgs.asahi-audio`, which gets rid of the DSP pipelines (incl. filters) from pipewire. `services.pipewire.extraLv2Packages` also doesn't do anything.

I guess you don't need asahi-audio to hear normal-volume audio. We could make the `speakersafetyd` part and the `$ALSA_CONFIG_UCM2` not conditional on `setupAsahiSound`, but just straight up removing asahi-audio is definitely not right.

ghost:
I've changed the proposal to remove the sound module itself and transition it into `hardware.asahi.enable` so that the sound components required for asahi are conditionally added if sound modules (in this case, pipewire) are enabled.

It also changed pulseaudio to `lib.mkForce` since we don't want users to be using pulseaudio if at all. 

I'm also not sure about users just using ALSA and whether speakersafetyd should be activated for them as well. 

The removal of `pkgs.asahi-audio` in the previous proposal is reverted since it doesn't allow the lv2 packages to be used.

sezaru:
@normalcea Just to let you know, after the latest changes to this branch, there seems to be a regression and now the sound is back to 100% after a reboot/logout.

dramforever:
> Just to let you know, after the latest changes to this branch, there seems to be a regression and now the sound is back to 100% after a reboot/logout.

This is sort of expected --- the original change in this PR was bogus and made under the assumption that asahi-audio was useless, which is clearly not true since Fedora Asahi continues to ship and update it, and asahi-audio itself was updated as well.

However, setting (effectively):

```
services.pipewire.configPackages = [ pkgs.asahi-audio ];
services.pipewire.wireplumber.configPackages = [ pkgs.asahi-audio ];
```

Causes the 100% volume thing, and this I still have no idea why.

sezaru:
I don't have the link anymore, but when I was researching about the issue, I do remember reading somewhere that adding the configs to pipewire and wireplumber were not needed anymore and it was fully removed from Fedora Asahi.

That's is also why asahi-diagnose complains with the default nix config.

yuyuyureka:
It is really confusing to me what kind of problem this PR is supposed to solve, and how it does it.
The diff is difficult to review because stuff is removed in one place and slightly different stuff is added in another place.

Can you please at first keep the sound module, and make any changes to the settings within the existing sound module? Then we can evaluate enabling setupAsahiSound by default later.

ghost:
Initially this was submitted to solve #352, but it turns out this was not the right approach, so the PR morphed into setting up the asahi sound components when the related sound modules are enabled and removes `setupAsahiSound` (thus the goal of making it implicit when sound is enabled)

Relooking over this PR, this change assumes that all mac models have both microphone and speaker support, which is not true, so I think this change is too awkward. 

I'll submit a new PR with changes to the existing sound module.

PR #362, opened 2025-10-05
Various updates (linux-asahi, u-boot-asahi)

yuyuyureka:


ghost:
I can vouch for the uboot-asahi and linux-asahi updates working correctly on my MBA M1 (2020). The timeout is indeed now 1 second instead of 5 in uboot and I haven't noticed any kernel regressions yet.

---

As for the m1n1 update, there is #349 which uses the m1n1 from nixpkgs which has the new version. I think that commit should be cherry picked onto this PR if we are doing an m1n1 update.

yuyuyureka:
Indeed your upstream m1n1 is working fine, so I removed the m1n1 update here and will prefer your PR instead — that is once nixos-unstable updates to a revision that doesn't break our cross ISO.

ghost:
I'd like to note that nixpkgs upstream m1n1 does *not* have chainloading nor any of the m1n1 tools that n-a-s's m1n1 has and I would definitely welcome either of those contributions upstream. 

chainloading is not required for our use case, but being able to compile it would be nice, tools is also in a similar boat.

PR #367, opened 2025-10-18
boot-m1n1: respect `boot.m1n1CustomLogo` again for `bootM1n1`

ghost:
This restores being able to set a custom logo.


yuyuyureka:
Thank you!

PR #368, opened 2025-10-20
docs: overhaul markdown docs into a texinfo manual

ghost:
Converts the entirety of the documentation into the texinfo format with adjustments made throughout to accommodate the new formalized structure.

The benefit of using texinfo is that we are able to write documentation in a structured format that can be read from as low as the TTY to a web browser in a desktop environment. It also allows for the documentation to be installed on the user's machine to be read locally at any time.

Currently, the documentation is presented through a load-bearing `uefi-standalone.md` that squeezes nearly everything together where it was hard to understand which information belonged to which section (for example, the nixos configuration recommendations being part of the install process). In addition, there was no clear markdown style guide which led to each markdown file being written inconsistently. Markdown also doesn't lead itself well to growing due to the lack of macros (basically no variables).

This also creates a new package called `manual` that builds texinfo manual that can be installed and read on the users system. 

There are a lot of changes made to the documentation, so I'm marking this as a draft for now (+ I want to square away the not ideal CSS file).

### DIFF
--- a/apple-silicon-support/packages/nas-manual/nixos-apple-silicon.texi
+++ b/apple-silicon-support/packages/nas-manual/nixos-apple-silicon.texi
@@ -4,6 +4,7 @@
 @set VERSION release-2025-11-18
 @set DEVELOPMENT
 @set FORGE_URL https://github.com/nix-community/nixos-apple-silicon
+@set MATRIX @url{https://matrix.to/#/%23asahi:nixos.org, asahi:nixos.org}
 @set MAINTAINERS @@tpwrules, @@flokli, @@yuyuyureka
 @set BUG_TRACKER @url{https://github.com/nix-community/nixos-apple-silicon/issues, Bug Tracker}
 @set RELEASES_MIRROR @url{https://github.com/nix-community/nixos-apple-silicon/releases, GitHub Releases}
@@ -65,6 +66,13 @@ The version of the manual you are reading is @value{VERSION}
 @quotation Note
 This is a development version of this manual and may not be accurate
 to the latest tagged release (@value{VERSION}).
+@end quotation
+@end ifset
+
+@ifset MATRIX
+@quotation Note
+Discussion regarding apple-silicon support in NixOS can be had at the
+dedicated matrix room located at (@value{MATRIX})
 @end quotation
 @end ifset
 


ghost:
New Changes:

- Added a CSS stylesheet to the HTML output, this presents the manual in a centered-sheet with styling inspiration taken from both NixOS and GitHub.
- Added the manual as a package output in both the flake as well as in the installer.
   - The package is essentially a script wrapper that allows for opening the package using `info` or your web browser using `start-manual-{info/web}` from the built manual
- Edited the README to mention that reading the manual is possible by building it with an existing installation of Nix, while this does take away from the convenience of reading the docs from GitHub's own viewer, I think the benefits of being able to allow users to build versioned manuals as well as removing the reliance on GitHub in this aspect is valuable.
- In addition, made the default package the manual, this will cause the package to be built in shells but it only takes a couple of seconds to do so. This change is more or less revertable.

This is a fundamental overhaul on how documentation is done, but I believe that long-term it will allow for the project to be more flexible in terms of adding more chapters to documentation in a maintainable way (also allows for translations!)

To test/read from a checkout of this PR (I've also tested that this is accessible from the installer ISO)

```shellsession
$ nix build .#nas-manual -o manual
$ ./manual/bin/start-manual-web
[Note]: Run this script with any positional arguments present for split manual
```

Or directly:

`nix run github:normalcea/nixos-apple-silicon/use-texinfo`

tpwrules:
Thanks for the effort, it is interesting and presents a lot of opportunities for the manual to evolve.

I have one big concern about your proposal, which is the apparent necessity of an existing computer with Nix to read the manual. I think it needs to be easily readable on a phone in the event of problems, it's easy to imagine a scenario where something breaks and another computer is not immediately available. We do already have versioned manuals by keeping track of the files in our repo.

For this reason, I think we do need a CI job to push a rendered copy of the manual somewhere with an easily accessible link. It could be GitHub pages or something else operated by nix-community. (I happen to know the proprietor of https://grebedoc.dev/ if independence from GitHub is a big concern)

ghost:
Force pushed to add an "apps" output to allow for building and viewing the manual without having to go through the cross supported systems in the "packages" output. This feels very hacky, but it's actually supported within flakes[^1]

This should aid users who just want to view the manual on any Nix supported system locally.

[^1]: https://nix.dev/manual/nix/2.28/command-ref/new-cli/nix3-run.html?highlight=apps#apps


Re: rendered copy of the manual @tpwrules

I agree, a pre-rendered copy would be necessary for smartphone users and in cases where a Nix installation is not handy nearby. I don't really mind where it's hosted and I'll defer to your judgment. GitHub Pages seems like the path of least resistance though since we are already hosted on GitHub.

edit (Nov. 11th, 2025): To clarify, the independence aspect is already handled by the fact that the manual is now a package output in the repository and so it can be installed on users' systems or modified/forked.

ghost:
Rebased on the current 2025-11-18 update, I also squashed a the integration part into a single commit that adds both the app and package output for the manual.

ghost:
Additional changes: 

- Got rid of adding prettier to the formatter because it pulls nodejs into the closure size and that isn't worth it. We can just check for clean CSS diffs.
- Added a commit handling "next" manuals to meet users who use nixos-apple-silicon as a rolling release and so need up-to-date manuals specific to them. Users who need specific versioned manuals for versioned installers can read them from the release-branch.
- Edit: Corrected README to not have the `$` in the one-liner to allow for copy-pasting.

@tpwrules 

I think these are the changes I'm comfortable with for re-review when you have the time. Also pinging for a follow-up of having a documentation page since that's needed if we're going to go with this approach.

ghost:
Force pushed and rebased on current master. I've simplified the manual package by not having a separate attribute for darwin platforms and I also added the tiny-dfr section that was recently merged.

@tpwrules what are your current thoughts on this proposal? it's been a while since we last spoke.

ghost:
I added a commit to link to the matrix room for n-a-s since it's difficult to discover that this room exists without word of mouth.

ghost:
Setting as draft because this needs a better solution for integration: https://github.com/nix-community/nixos-apple-silicon/pull/368/changes/267cf08dfccb4788b471f7797aa1daea198b5e34

flokli:
NAK. This is gonna make the documentation way less approachable and is a downgrade.

PR #377, opened 2025-11-09
.github: revamp installer CI

tpwrules:
Big things:
* now built on ARM runner so the kernel is available for the user
* split up to reduce problems with running out of space or time
* doesn't use nix-community cache or builders, avoiding contamination

Tested that it boots and installs okay.

Build succeeded on my fork: https://github.com/tpwrules/my-nixos-apple-silicon/actions/runs/19211475096

### DIFF
--- a/.github/workflows/installer.yml
+++ b/.github/workflows/installer.yml
@@ -50,7 +50,10 @@ jobs:
         run: |
           # rest of the installer is downloaded and agglomerated and a few minor things are built
           nix build --extra-experimental-features 'nix-command flakes' .#installer-bootstrap -j2 -L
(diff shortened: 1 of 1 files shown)


yuyuyureka:
Couldn't this be simplified a lot by only unifying cache.yml and installer.yml? cache.yml already builds the installer images, it would just need an additional step to upload them somewhere when the pipeline is run on a release tag.

tpwrules:
This avoids any risk of compromise from the runners or cache service. It’s well within our time and space budgets now and is used infrequently so I think it’s better this way.

Maybe if we publish them more often it would be better to move.

Issue #361, opened 2025-10-02
installer ISO does not boot on 2020 Mac Mini M1


ebrensi:
On `main` branch, latest commit (`db555f5ac04144c5a56d60dbf4e15a3e58ccae08`) as of 2025/10/02.

The installer ISO created with `nix build .#installer-bootstrap -o installer -j4 -L` does not boot.  I don't know enough about the boot process to say what is failing.  It does get past the boot menu and Stage 1 of the boot process. It has some kind of error and hangs on Stage 2.

The pre-built image works fine.

yuyuyureka:
I believe this will fix it: https://github.com/AsahiLinux/asahi-installer-data/pull/1

jannau:
The PR is merged. For now the updated boot.bin is only in the testing installer: `curl -L https://alx.sh/dev | sh`

tpwrules:
The official installer now has the update as well.

PR #378, opened 2025-11-09
release 2025-11-18

tpwrules:
On top of #376 and #377 . Will rebase once those are merged.

Updates nixpkgs and fixes an issue causing stage1 errors especially in the installer.

I am still having intermittent DHCP issues in the installer but this is quite probably my home network and they seem to have abated so whatever.

We probably need to wait for Asahi to update the official installer so that this image will work on a fresh setup before releasing it. See #361 .

Will fully test after rebase.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -2,6 +2,21 @@
 
 This file contains important information for each release.
 
+## 2025-11-18
+
+This release drops every custom package except the kernel and U-Boot. These
+cannot be dropped due to upstream nixpkgs' policy of not accepting any more
+custom versions of these packages. Many thanks to contributors who worked in
+this repository and with nixpkgs to achieve this goal.
+
+THe kernel build is now also a full NixOS kernel build, enabling every module.
+This means no more issues with hardware support and a more standard experience.
+This does, however, increase build times, but CI and a public cache (courtesy of
+nix-community and Cachix) have been set up to alleviate this.
+
+Installer images are now built for ARM by default using public GitHub Actions
+runners.
+
 ## 2025-08-23
 
 This release drops our custom Mesa override, which fixes a build issue with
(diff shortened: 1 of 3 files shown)


tpwrules:
Need to update the docs with the U-Boot instructions [from Gentoo](https://wiki.gentoo.org/wiki/Project:Asahi/Guide#Boot_into_the_bootable_installation_media) for the Asahi U-Boot without bootmenu:

```
# setenv boot_targets "usb"
# setenv bootmeths "efi"
# boot
```

tpwrules:
Testing is going well, still some documentation work to do. I plan to hold back nixpkgs a bit and add a warning due to reported problems with Mesa 25.3.0.

docs/release-notes.md
```diff
@@ -2,6 +2,21 @@
 
 This file contains important information for each release.
 
+## 2025-11-xx
+
+This release drops every custom package except the kernel and U-Boot. These
+cannot be dropped due to upstream nixpkgs' policy of not increasing the number
+of same. Many thanks to contributors who worked in this repository and with
```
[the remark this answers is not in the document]
tpwrules: This is an idiomatic and evidently hard to understand expression. I have cleaned it up. Thanks for testing.

tpwrules:
Tested on 16" MacBook Pro M1 Max that the installer works (including with the new Asahi Linux official installer binaries) both native and cross, that the system boots and seems stable, that networking and graphics and sound are all functional.

No notes except for the new Mesa warning. Added a couple new items to the docs as well (yes I know they are getting unwieldy).

apple-silicon-support/modules/default.nix
```diff
@@ -28,6 +28,12 @@
           }
         else
           pkgs;
+
+      warnings = lib.optional (lib.versionAtLeast pkgs.mesa.version "25.3") ''
```
dramforever: There really should be a way to silence this warning

tpwrules: Thanks, good idea. Hopefully this won't stay around for long, if it does we may add our own pin.

PR #382, opened 2025-11-24
installer-bootstrap: remove unneeded doc options

ghost:
We import 3 files in the bootstrap-installer 

``` nix
  imports = [
    (modulesPath + "/profiles/minimal.nix")
    (modulesPath + "/profiles/installation-device.nix")
    (modulesPath + "/installer/cd-dvd/iso-image.nix")
  ];
```

In  `nixos/modules/profiles/minimal.nix`

``` nix
  documentation = {
    enable = mkDefault false;
    doc.enable = mkDefault false;
    info.enable = mkDefault false;
    man.enable = mkDefault false;
    nixos.enable = mkDefault false;
  };
```

We had duplicated this in our configuration, so we can safely remove it.

In `nixos/modules/profiles/installation-device.nix`

``` nix
    # Provide networkmanager for easy network configuration.
    networking.networkmanager.enable = true;
```

NetworkManager is used by default in the minimal ISO and in my initial testing, connecting to the network works fine but requires the user to use sudo or be root which I can't find the root cause of.

The ARM-compiled installer gets an increase from the networkmanager addition from 529MB -> 546MB (~17MB) on my testing. I have not checked the difference on the cross-compilation.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -5,6 +5,9 @@
 #
 
 # treewide: re-format
+e0ca5759fae2b906785799a2c83cdba267e82693
+
+# treewide: re-format
 69532003a7d4b95c0fc508c6ee777411a9b3dd3b
 
 # treewide: reformat with nixfmt v1.0.0


ghost:
Removed the iwd change and just left in the doc options removal and the treewide re-format.

flokli:
I built the installer image successfully (using cross).
Looking forward to the iwd > NetworkManager PR, alongside with the closure comparison!

PR #386, opened 2025-11-29
linux-asahi: 6.17.7-2 -> 6.17.9-1

yuyuyureka:


karlskewes (commented):
The installer built with 6.17.9 boots more reliably on my M2 Pro mini than 6.17.7. No immediate graphics glitches, no need to disable mode setting.
Also combined with the recent alx.sh update.
Have not tested installed on system yet (btrfs issue).

karlskewes (approved):
Booted from USB stick and installed fine on my M2 Pro Mac Mini.

PR #387, opened 2025-12-09
apple-silicon-support: add drirc to work-around website blocking

ghost:
Some websites use a third-party security API which blanket blocks asahi machines. This setting has already been merged upstream (just 11 hours ago at the time of writing) but that will have to wait until the next Mesa update.

In the meantime, this snippet can be used to apply the patch to the existing configuration. 

Credits to @waltmck  for the implementation. My addition was changing the source FOD to fedora pagure and using `drirc.d` instead of `drirc` to avoid potential clobbering.

Context:
- https://pagure.io/fedora-asahi/asahi-platform-metapackage/pull-request/25

### DIFF
--- a/apple-silicon-support/modules/10-asahi-browser-apple.conf
+++ b/apple-silicon-support/modules/10-asahi-browser-apple.conf
@@ -1,0 +1,92 @@
+<?xml version="1.0" standalone="yes"?>
+<!--
+
+============================================
+Application bugs worked around in this file:
+============================================
+
+* web broswer OpenGL renderer override fo asahi
+  Several web sites (paypal.com, etsy.com) apparently block on "Apple" in
+  WebGL renderer strings when the UA reports Linux as OS.
+  See https://github.com/webcompat/web-bugs/issues/189524
+
+-->
+
+<!DOCTYPE driconf [
+   <!ELEMENT driconf      (device+)>
+   <!ELEMENT device       (application | engine)+>
+   <!ATTLIST device       driver CDATA #IMPLIED
+                          device CDATA #IMPLIED>
+   <!ELEMENT application  (option+)>
+   <!ATTLIST application  name CDATA #REQUIRED
+                          executable CDATA #IMPLIED
+                          executable_regexp CDATA #IMPLIED
+                          sha1 CDATA #IMPLIED
+                          application_name_match CDATA #IMPLIED
+                          application_versions CDATA #IMPLIED>
+   <!ELEMENT engine       (option+)>
+
+   <!-- engine_name_match: A regexp matching the engine name -->
+   <!-- engine_versions: A version in range format
+             (version 1 to 4 : "1:4") -->
+
+   <!ATTLIST engine       engine_name_match CDATA #REQUIRED
+                          engine_versions CDATA #IMPLIED>
+
+   <!ELEMENT option       EMPTY>
+   <!ATTLIST option       name CDATA #REQUIRED
+                          value CDATA #REQUIRED>
+]>
+
+<driconf>
+    <device driver="asahi">
+        <application name="Firefox" executable="firefox">
+            <option name="force_gl_renderer" value="AGX G1Nx"/>
+        </application>
+        <application name="Firefox Wrapped" executable=".firefox-wrapped">
+            <option name="force_gl_renderer" value="AGX G1Nx"/>
+        </application>
+        <application name="Firefox Beta Wrapped" executable=".firefox-beta-wrapped">
+            <option name="force_gl_renderer" value="AGX G1Nx"/>
+        </application>
+        <application name="Firefox ESR Wrapped" executable=".firefox-esr-wrapped">
+            <option name="force_gl_renderer" value="AGX G1Nx"/>
+        </application>
+        <application name="Firefox ESR" executable="firefox-esr">
+            <option name="force_gl_renderer" value="AGX G1Nx"/>
+        </application>
+        <application name="Firefox Beta" executable="firefox-beta">
+            <option name="force_gl_renderer" value="AGX G1Nx"/>
+        </application>
+        <application name="Chromium" executable="chromium">
+            <option name="force_gl_renderer" value="AGX G1Nx"/>
+        </application>
+        <application name="Chromium" executable="chromium-browser">
+            <option name="force_gl_renderer" value="AGX G1Nx"/>
+        </application>
+        <application name="Google Chrome" executable="chrome">
+            <option name="force_gl_renderer" value="AGX G1Nx"/>
+        </application>
+        <application name="vivaldi-bin" executable="vivaldi-bin">
+            <option name="force_gl_renderer" value="AGX G1Nx"/>
+        </application>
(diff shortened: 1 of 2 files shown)


psanford:
Thanks for this patch.

On my nixos system, this config does not get picked up when using:
```
   environment.etc."drirc.d/10-asahi-browser-apple.conf".source = pkgs.fetchurl {
```

However it does work correctly with:
```
environment.etc."drirc".source = pkgs.fetchurl {
```

EDIT:
I'm testing this in chromium with the following javascript:
```
const canvas = document.createElement('canvas')
const gl = canvas.getContext('webgl');
const debugInfo = gl.getExtension('WEBGL_debug_renderer_info');
gl.getParameter(debugInfo.UNMASKED_RENDERER_WEBGL)
```

waltmck:
> Thanks for this patch.
> 
> On my nixos system, this config does not get picked up when using:
> 
> ```
>    environment.etc."drirc.d/10-asahi-browser-apple.conf".source = pkgs.fetchurl {
> ```
> 
> However it does work correctly with:
> 
> ```
> environment.etc."drirc".source = pkgs.fetchurl {
> ```
> 
> EDIT: I'm testing this in chromium with the following javascript:
> 
> ```
> const canvas = document.createElement('canvas')
> const gl = canvas.getContext('webgl');
> const debugInfo = gl.getExtension('WEBGL_debug_renderer_info');
> gl.getParameter(debugInfo.UNMASKED_RENDERER_WEBGL)
> ```

Yep, credit to K900 for walking me through that part. It is also necessary to add `.firefox-wrapped`, `.firefox-beta-wrapped`, `.firefox-esr-wrapped` to the driconf file due to the way those browsers are packaged on NixOS. I have a complete working solution in [my gist](https://gist.github.com/waltmck/6edce28d2f7abc356f8f04b2154e2bb8).

An easy way to test whether the driconf is being applied on firefox is to go to about:support and search for `AGX G1Nx`.

ghost:
Changed to just track the modified configuration file and read it from there, we'll probably need to make this solution more robust since it just tracks a few popular web browsers (not all of them) and needs to also be patched for nix binaries. 

My mistake for just checking with chromium and assuming the lack of security blocks meant it worked on FF.

ghost:
I've implemented it via a module option as well in a spin-off branch https://github.com/normalcea/nixos-apple-silicon/tree/mesa-web-captcha-fix-option

If this solution is better than I think we can go with a module option instead and not have to track a static conf file in-tree.

yuyuyureka:
From what I understand, upstream mesa will add a workaround in 25.3.x?

The -wrapped thing should be handled in nixpkgs too, so that other firefox/chromium workarounds are applied too.

dramforever:
This should be no longer necessary on 25.3.2

waltmck:
> This should be no longer necessary on 25.3.2

I think that it will still be necessary until Nixpkgs patches Mesa to add `.*-wrapped` to the list of known browsers in the driconf.

ghost:
@waltmck To clarify, do the original browser names have to still be included or is it possible to replace the application executable names with just the `.*-wrapped` version and discard the original tags?

waltmck:
> @waltmck To clarify, do the original browser names have to still be included or is it possible to replace the application executable names with just the `.*-wrapped` version and discard the original tags?

Isn't it possible that third-party-packaged browsers could link against nixpkgs' Mesa? If so, we would also need to keep the standard names to support this use-case. There is no downside to keeping them, so it might be smart to do so just to be on the safe side.

ghost:
Can we still confirm that this is an issue after 25.3.2? I seem to be unable to trigger it on my end after removing the overrides. WSJ correctly blocks my VPN proxy but then allows me to enter after I turn it off, only giving me a puzzle captcha. My driver renderer is still showing "Mesa -- Apple M1 (G13G B1)"

On GNOME web and Firefox I'm able to register for an etsy account and also go into the login screen on paypal and also access WSJ. On ungoogled chromium I can do much the same thing.

I think the security module these sites were using either turned off the checks on the GPU, didn't turn them off but didn't alert on apple linux, or used a captcha in the case of WSJ.

yuyuyureka:
I'm still blocked from etsy.com but not wsj

flokli:
> > This should be no longer necessary on 25.3.2
> 
> I think that it will still be necessary until Nixpkgs patches Mesa to add `.*-wrapped` to the list of known browsers in the driconf.

Where is `executable` read from? Is it `argv0`? In that case, we could just use `--inherit-argv0` in our `makeWrapper` calls in our browser derivations (and some might already do, or not use a wrapper, so the list needs a bit of manual checking).

If we can't easily make the upstream driconf work for us that way:

I peeked at `src/util/00-mesa-defaults.conf` from mesa src, changed in https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/38845/diffs - there's not only an `executable` attribute, but various other fields available to do the matching.

We could play around with these - `application_name_match` might be a better candidate that could work on all distros? This should even be upstream-able to mesa directly.

If this doesn't work either, and we need to patch the driconf, it should be a custom patch in the nixpkgs mesa derivation, not here. `driconf` not applying to our browsers is not really something we need to ship here.

ghost:
Re: @flokli 
> If this doesn't work either, and we need to patch the driconf, it should be a custom patch in the nixpkgs mesa derivation, not here. driconf not applying to our browsers is not really something we need to ship here.

I have a WIP branch that patched the driconf programmatically via xpath substitution using xmlstarlet (https://github.com/NixOS/nixpkgs/compare/master...normalcea:nixpkgs:mesa-patch-asahi-overrides) I hadn't submitted it to nixpkgs since I was unsure of it was the best approach (and also some sites affected magically deciding to start working and then stop working).

> We could play around with these - application_name_match might be a better candidate that could work on all distros? This should even be upstream-able to mesa directly.

This would be the ideal solution, but maybe there's a reason why `executable` was chosen and not it by upstream? There'd need to be more testing done on this.

---
I also agree it's not something worth shipping here unless as a last resort. I haven't rescinded the PR for visibility reasons.

flokli:
Changing the driconf is something I'd ideally not do at all, that's why I suggested looking at the wrappers (and whether `--inherit-argv0` would help first.

I tried epiphany and chromium, and when opening `chrome://gpu` they both list "AGX G13/G14" as a substring in the `GL_RENDERER` variable. So I'd guess everything already works properly for these browsers.

So I'd guess the problem isn't that widespread, and we shouldn't just blindly duplicate `-wrapped` rules for every browser we don't even need (which this PR and the linked nixpkgs branch do).

Firefox does indeed seem to need fixing, it still shows "Apple M2" in WebGL {1,2} Driver renderer" at `about:support`, and it has a few wrappers. I really think we should try going down the route of either fixing this in the wrappers, or if that's not sufficient, tweaking the build expressions to ensure the binaries are named the way they are on other distros.

flokli:
Closing this PR as it's nothing we'd ship as-is, I'll open a nixpkgs tracking issue

flokli:
Let's continue over in https://github.com/NixOS/nixpkgs/issues/476721.

PR #397, opened 2025-12-29
treewide: remove internal dependency on flakes

ghost:
This patch series removes flakes tree-wide and in its place uses a JSON lockfile produced by [npins](https://github.com/andir/npins) to pin various nixpkgs versions.

The benefit of this approach is that it does not force the project to rely on flakes for development and distribution as the format exposes many issues:

- It's incredibly difficult to mix cross-compiled packages and non-cross packages due to the systems attribute prefixing `packages`. Trying to work around this means not following the flakes schema. I encountered this when trying to compile the documentation into a manual package which shouldn't be placed in cross but can't also be placed outside of it, leading to awkward workarounds (obligatory "flakes breaks cross" comment, which is pressing since cross-comp is a main component of n-a-s).
- Flakes are considered experimental, this means that there is no formal contract for using them. Many projects using Nix are already de-facto using flakes despite the fact the fact that the feature is not stable (and hasn't been for many years and probably never will be). This has led to an "informal economy" of sorts around flakes with packages such as "flake-compat" and "flake-utils" springing up as well as needing to preamble nix3 and flake commands with `--extra-experimental-features` It's simultaneously necessary to augment flakes with shibboleths like `forEachSystem` but also a headache to break away from the flake structure at any point.
- Pure evaluation (evaluation from the nix store) gives us no discernible benefits. In fact, it's not even conducive to using `asahi-firmware` since flakes requires all components to be tracked by git in order to be used. There's a hack to do this with `pkgs.requireFile` that's not documented.
- Flakes are overly complicated and have always been. I argue my npins-based replacement is far simpler and more idiomatic to Nixpkgs.

In converting to npins, I tried to keep the attributes more or less in line with the flake nomenclature as it's a legitimately good part of flakes to have standardized project outputs. I did move the `installer-bootstrap` derivation out of `packages` and into its own attribute since it's not a package to install and we have no arbitrary restriction placed on us by the flakes schema.

I also modified the GitHub Actions CI to the best of my ability. However, I haven't been able to test the changes to the `cache.yml` workflow outside of local testing, I have tested the `installer.yml` workflow via my fork. More testing should be done on making sure the CI transitions properly. I've also updated the documentation to replace usages of flakes (as well as the already deprecated section on building m1n1).

In its place, a replacement flake.nix is used so that downstream flake users don't need a manual change on their end. This makes it so that we still inherit the project publishing powers of flakes but none of the additional baggage that comes with adopting flakes as an entire paradigm.

### DIFF
--- a/update-inputs.sh
+++ b/update-inputs.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env nix-shell
+#!nix-shell -i bash -p jq
+# shellcheck shell=bash
+
+set -euo pipefail
+
+readonly NPINS_VERSION="0.3.1"
+npins_version_string=$(npins --version)
+
+if [[ ! "${npins_version_string}" == *"${NPINS_VERSION}" ]]; then
+    printf "Required 'npins %s', found '%s'\n" \
+           "${NPINS_VERSION}" \
+           "${npins_version_string}" >&2
+    printf "Run from project's 'nix-shell' to ensure versions match!\n" >&2
+    exit 1
+fi
+
+if npins --lock-file npins.json update; then
+    readarray -t nixpkgs <<< "$(jq -r '.pins."nixos-unstable" | .revision, .hash' npins.json)"
+    revision="${nixpkgs[0]}"
+    sri_hash=$(nix-hash --to-sri --type sha256 "${nixpkgs[1]}")
+    jq --arg jq_rev "${revision}" --arg jq_hash "${sri_hash}" \
+       '.nodes.nixpkgs.locked.rev |= $jq_rev | .nodes.nixpkgs.locked.narHash |= $jq_hash' \
+       flake.lock > flake.lock.tmp
+    mv flake.lock.tmp flake.lock
+fi


ghost:
Added two additional commits:

- Adds back the flake as a shim (an API indirection) that imports n-a-s's default.nix with the "nixpkgs" input. This should be a 1-to-1 compatibility layer with the previous flake file and existing downstream flake users shouldn't suffer a regression that needs manual intervention (unless they for some reason use the `devShells`, `formatter`, `checks` output in their downstream flakes).
- Creates a condition in which if the host platform is the same as the target platform (`"aarch64-linux"`), then just use the existing `pkgs` attribute instead of re-importing. This eliminates any chance of duplicate nixpkgs evaluations across any Nix evaluator.

flokli:
As much as I dislike the overhead and verbosity the flake interface brings, there's people consuming this as a flake, including ourselves for the nix-community buildbot, so I don't see this feasible, at least not in the current state. Downgrading to draft.

ghost:
The buildbot fails as the previous flake used didn't expose a `checks` output attribute. All the buildbot seems to do is build the `checks` output recursively (I can't remember if this is related to the concerns brought up here but it looks like it does https://github.com/nix-community/nixos-apple-silicon/pull/341#issuecomment-3314883411)

This can be replaced by both `treefmt --ci` (provided by nixfmt-tree and errors (exits 1) on change in reformatting) and `nix-eval-jobs --no-instantiate --force-recurse default.nix` (which will fail on any invalid derivation, but not build them). 

Both of these would not require needing the nix-community buildbot for CI (we can reuse GHA for this). I've added the checks and formatter outputs back into the flake though in the meantime. It should also be possible to change the nix-community CI to do what I mentioned above if that's desired.

> there's people consuming this as a flake,

Downstream flake users on Asahi machines should be covered by the kept flake other than not having cross-comp (which is not even considered an issue by some flake contingents[^1]) and has been stalling for a long while for official support[^2]. 

Flakes are fine as an extension format to Nix for publishing projects (and are indeed very convenient in that way), but stable nix should be our baseline until flakes is part of stable nix. We only need one "flake feature" for n-a-s which is tracking nixpkgs revisions to detect regressions and to give the last good known nixpkgs revision when those regressions are spotted to downstream users.

---

The motivation for this change also comes from when I was trying to build the docs into a package (firstly as mdbook and then second as texinfo). I realized that mixing non-cross comp and cross-comp packages was out of my reach since we re-import nixpkgs to export the cross-comp to both x86 and aarch64 system attributes, thus merging the attrsets was not possible or at the very least intuitive to do. This limitation could not just be for documentation but for other packages for asahi that could be built for the host platform other than aarch64. 

[^1]: https://determinate.systems/blog/experimental-does-not-mean-unstable/
[^2]: https://github.com/NixOS/nix/issues/5157

ghost:
I've modified the PR title to be clearer on what my actionable goal is, which is to follow nixpkgs' approach: expose a flake entrypoint to end users but internally are able to come up with approaches that are not constrained by flake's current limitations. "Remove flakes" wasn't clear at all on what I intended and what this eventually developed to:

Blockers I've personally noted so far:

- [ ] nix-buildbot builds the "checks" output of the flake, it should either be modified to not do this or we keep the checks output in the flake we have and take no action. We can also replace it with GitHub Actions and not rely on it at all since the linting it does is simple. 
- [ ] cache.yml uses `nix-fast-build` which has a hard dependency on flakes because of remote builds (I'm not sure if that's relevant to us). I think the approach here is to replace its functionality since we only build 3 packages (uboot, linux) -> installer-bootstrap (cross and host). We can use `nix-eval-jobs` into `jq` to grab the outpaths we need to upload to cachix. 
- [ ] installer.yml: Since we don't use flakes to build the installer, the name of the iso is different than what flakes uses, this needs to be verified that the branch name can be substituted.
- [ ] Updating the `npins.json` file and the `flake.lock` is not synchronized nor do they share the same inputs.
- [ ] npins is at 0.3.1 with the pinned `nixos-unstable` used currently, either have to downgrade or wait for the update at (https://github.com/NixOS/nixpkgs/commit/373455a414f64b066ace1fe09d57bc96f5a6e5a1)

ghost:
Re @flokli (TL;DR) did more work on this, I think this merits a deeper look. 

I've done more work on this, I think what I have now is feasible and will be a net gain over the previous pure flakes approach.

- As for the nix-buildbot, the checks output is preserved and acts the same as it did previously, if changing the buildbot is possible, then here is an example of a linting action that could be used, otherwise the buildbot can be kept the same or be replaced with GHA.

Example of linting action that is as space efficient as possible:
``` yaml
name: lint

on:
  pull_request:
jobs:
  linters:
    name: Linters
    runs-on: ubuntu-24.04
    steps:
      - uses: actions/checkout@v6
      - uses: cachix/install-nix-action@v31

      - name: Check formatting
        run: |
             formatter_storepath=$(nix-build -A formatter)
             "${formatter_storepath}"/bin/treefmt --ci

      - name: Check derivations (do not build)
        run: nix-instantiate -A packages -A installer-bootstrap -A formatter -A shell
```

- Regarding cache.yml, I've updated it to simply use `nix-build` since we only have to evaluate three package derivations before building them. The workflow updates each nixpkgs pin to their latest version and then forks a matrix of each build with `--argstr nixpkgsBranch` controlling which nixpkgs is used. I have not tested pushing this configuration to a cachix instance yet but from reading the documentation[^1] this seems to be all you would need.

- installer.yml is kept more or less the same, the ISO output uses the name `pre-git` instead of the git SHA (as we use the GitHub archive instead of the Nix channel) but that doesn't affect the workflow being able to substitute the release name into the final ISO.

- As for the npins tooling, I've created a versioned `update-inputs.sh` shell script that updates all the pinned nixpkgs versions and also replaces the narHash and revision of the nixpkgs input in the `flake.lock` so they should always be synchronized. I also made it so that the npins lockfile follows the version and schema of npins available in nixpkgs and so downgraded the npins lockfile to 0.3. The only difference between 0.3 and 0.4 is that 0.4 uses SRI hashes in the lockfile.

No change has been made to the default nixos-unstable pinned revision, but additional nixpkgs branches have been added to the lockfile.

---

The default experience for downstream flakes consumers does not change with this rework. The only thing lost is cross-compilation which was never a supported path in flakes primitives anyway and it's best to just wait for flakes to catch up to this instead of us catching up to flakes.

[^1]: https://nix.dev/guides/recipes/continuous-integration-github-actions

flokli:
While I sympathize with some of the paint points mentioned here, there's no obligation whatsoever to use the flake entrypoint.

In my config, I simply import the `apple-silicon-support` subdir, similarly to how it's documented in `docs/uefi-standalone.md` (except my sources come from npins):

```
    ("${sources.nixos-apple-silicon}" + "/apple-silicon-support")
```

There's no documentation on how to use the flake entrypoint, and this PR also doesn't delete it. I don't see it improving the current situation.

PR #416, opened 2026-01-17
systemd-boot: remove forced `consoleMode = "0"`

vidhanio:
Setting `consoleMode = lib.mkForce "max"` seemed to work perfectly fine, and this comment/line was added in 2022, which means u-boot has possibly been updated since then to allow for this.

### DIFF
--- a/apple-silicon-support/modules/kernel/default.nix
+++ b/apple-silicon-support/modules/kernel/default.nix
@@ -76,9 +76,6 @@
     # U-Boot does not support EFI variables
     boot.loader.efi.canTouchEfiVariables = lib.mkForce false;
 
-    # U-Boot does not support switching console mode
-    boot.loader.systemd-boot.consoleMode = "0";
-
     # GRUB has to be installed as removable if the user chooses to use it
     boot.loader.grub = lib.mkDefault {
       efiSupport = true;


zzywysm:
This change was made to fix a specific boot error.

https://github.com/nix-community/nixos-apple-silicon/commit/7f0e3ce9dbf4366a9f73cf81dfbbfd345dfdffc2

Did you confirm that your change doesn’t reintroduce the error?

vidhanio:
searching through my journalctl finds no such error, and I don't see the mentioned red text anytime during my boot process either.

zzywysm:
Do you receive the error if you run these commands and then reboot your Mac?

sudo nix-collect-garbage --delete-old
sudo nix-collect-garbage -d
sudo /run/current-system/bin/switch-to-configuration boot

vidhanio:
Done, no errors came up.

vidhanio:
@zzywysm are there any blockers on this?

zzywysm:
Weren’t you setting it to max before?  Now you just don’t set it at all?

vidhanio:
@zzywysm i still have `lib.mkForce "max"` set, if that's what you're asking? Both of my test results were with that set as well

yuyuyureka (approved):
LGTM

PR #425, opened 2026-02-17
linux-asahi: 6.18.9-1 -> 6.18.10-1

yuyuyureka:


apple-silicon-support/packages/linux-asahi/default.nix
flokli: This needs updating too:

> Error: modDirVersion 6.18.9 specified in the Nix expression is wrong, it should be: 6.18.10

amiuhle:
I've been running it since 2026-02-17 18:21:13, seems to be fine

Issue #414, opened 2026-01-15
cut a new release


flokli:
There's been a few stability issues with USB, u-boot, m1n1 etc reported, for people starting with the latest release.

We should cut a new release.

Things still to be done by then

 - go through PRs since the last release, update release notes
 - #408 
 - #363

yuyuyureka:
This is blocked on the uefi-standalone payload on alx.sh being updated, as the current one ships device trees incompatible with the 6.18 USB-C implementation.

https://oftc.catirclogs.org/asahi-dev/2026-01-15#34963637;

zzywysm:
> This is blocked on the uefi-standalone payload on alx.sh being updated, as the current one ships device trees incompatible with the 6.18 USB-C implementation.

This is no longer blocked as of 20 minutes ago.

zzywysm:
Also note that there is a new 6.18.x kernel tag that should get rolled into the new release:

https://github.com/AsahiLinux/linux/releases/tag/asahi-6.18.15-1

zzywysm:
And there’s a new u-boot tag from January too

flokli:
@zzywysm you want to open individual PRs and do some testing? ;-)

yuyuyureka:
✔️ Re-created my UEFI standalone partition with a fresh copy from alx.sh and booted the NixOS ISO (linux-asahi 6.19.9-1) with it.

yuyuyureka:
I would like to update nixpkgs once nixos-unstable gets bumped, and enable NetworkManager by default to unify the installation steps with upstream NixOS.

zzywysm:
> [@zzywysm](https://github.com/zzywysm) you want to open individual PRs and do some testing? ;-)

Nah, it’s way more amusing to sit back and watch the official maintainers of this project merge in unreviewed, untested commits, and then do bare commits to the repository (i.e. without a PR) to fix their mistake.

https://github.com/nix-community/nixos-apple-silicon/pull/437
https://github.com/AsahiLinux/linux/releases/tag/asahi-6.19.9-2
https://github.com/nix-community/nixos-apple-silicon/commit/d0348f9b2fd00e91f818c44da2b643148186eaef

In the past an official maintainer called my PRs “opinionated”.  Surely opinionated and tested is better than broken and untested?

yuyuyureka:
[...]

PR #433, opened 2026-03-11
docs/uefi-standalone, installer: enable Network Manager, and make it use IWD for wireless connectivity

WillemToorenburgh:
This PR has evolved significantly since I originally created it. What started as me adding some info I found regarding DNS and IWD, is now enabling Network Manager in the installer and making it work with IWD. This brings the installer and documentation closer in-line with the standard NixOS installer. 

---

Hey there!

## Summary

This PR adds a few lines to `docs/uefi-standalone.md` regarding making IWD, NetworkManager, and the user's DNS system of choice, all correctly connect with each other.

It also implements these recommendations in `iso-configuration/installer-configuration.nix`.

## Context

I was having a bunch of trouble getting my machine to use an alternate DNS server, with no success using Plasma Settings via Network Manager or using something like `networking.nameservers = [ "10.0.0.6" ];`. Finally, after much troubleshooting, I discovered that [IWD uses `systemd-resolved`](https://man.archlinux.org/man/iwd.config.5#Network) by default, while [NixOS uses `resolvconf`](https://nixos.org/manual/nixos/stable/options#opt-networking.resolvconf.enable) by default. After setting IWD to use resolvconf, I was finally able to set my alternate DNS server in Plasma Settings.

## Future work

It would make sense for NetworkManager to automatically use IWD as its wireless backend when enabled, and for IWD to automatically use the user's configured DNS client (if compatible, and asserting when not), but that is work that should be done in nixpkgs.

### DIFF
--- a/iso-configuration/installer-configuration.nix
+++ b/iso-configuration/installer-configuration.nix
@@ -85,28 +85,15 @@
   hardware.asahi.setupAsahiSound = false;
   system.extraDependencies = lib.mkForce [ ];
 
-  # Disable wpa_supplicant because it can't use WPA3-SAE on broadcom chips that are used on macs and it is harder to use and less mainained than iwd in general
-  networking.wireless.enable = false;
-  # Enable iwd
+  # Enable iwd and make NetworkManager, enabled by default, use iwd for wifi
+  # We use iwd instead of wpa_supplicant because it can't use WPA3-SAE on broadcom chips that are used on macs,
+  # and it is harder to use and less mainained than iwd in general
+  networking.networkmanager.wifi.backend = "iwd";
+
   networking.wireless.iwd = {
-    enable = true;
-    settings.General.EnableNetworkConfiguration = true;
     # NixOS by default uses resolvconf for DNS, but iwd's default is systemd-resolved. Make iwd use resolvconf
     settings.Network.NameResolvingService = "resolvconf";
   };
-  # Make sure NetworkManager, also enabled by default, uses iwd for wifi
-  networking.networkmanager.wifi.backend = "iwd";
-
-  # let user know to use iwctl to get access to iwd
-  services.getty.helpLine = lib.mkForce ''
-    The "nixos" and "root" accounts have empty passwords.
-
-    To log in over ssh you must set a password for either "nixos" or "root"
-    with `passwd` (prefix with `sudo` for "root"), or add your public key to
-    /home/nixos/.ssh/authorized_keys or /root/.ssh/authorized_keys.
-
-    To set up a wireless connection, run `iwctl`.
-  '';
 
   nixpkgs.overlays =
     lib.optionals (config.nixpkgs.hostPlatform.system != config.nixpkgs.buildPlatform.system)
--- a/iso-configuration/installer-configuration.nix
+++ b/iso-configuration/installer-configuration.nix
@@ -90,11 +90,6 @@
   # and it is harder to use and less mainained than iwd in general
   networking.networkmanager.wifi.backend = "iwd";
 
-  networking.wireless.iwd = {
-    # NixOS by default uses resolvconf for DNS, but iwd's default is systemd-resolved. Make iwd use resolvconf
-    settings.Network.NameResolvingService = "resolvconf";
-  };
-
   nixpkgs.overlays =
     lib.optionals (config.nixpkgs.hostPlatform.system != config.nixpkgs.buildPlatform.system)
       [
--- a/iso-configuration/installer-configuration.nix
+++ b/iso-configuration/installer-configuration.nix
@@ -91,8 +91,11 @@
   networking.wireless.iwd = {
     enable = true;
     settings.General.EnableNetworkConfiguration = true;
+    # NixOS by default uses resolvconf for DNS, but iwd's default is systemd-resolved. Make iwd use resolvconf
+    settings.Network.NameResolvingService = "resolvconf";
   };
-  networking.networkmanager.enable = lib.mkForce false;
+  # Make sure NetworkManager, also enabled by default, uses iwd for wifi
+  networking.networkmanager.wifi.backend = "iwd";
 
   # let user know to use iwctl to get access to iwd
   services.getty.helpLine = lib.mkForce ''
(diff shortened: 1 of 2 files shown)


flokli:
I spoke with @yuyuyureka about this a bit. It might be less surprising if we'd enable NetworkManager in the installer (to match what the official NixOS installer does), and only document that it should be used with the iwd backend.

That way interactive dns configuration should also work, and it's a combination of tools that's way more commonly used on other NixOS systems as well.

WDYT?

WillemToorenburgh:
OH! I think I might have put the new doc lines in the wrong area. The problem was not trying to use my DNS server in the installer environment, it was _after_ the full system installation. The guidance I'm suggesting isn't inherently Apple silicon installation related, but more a consequence of recommending IWD. It just felt more correct to put that information as close to enabling IWD as possible as it was the major shakeup for me in comparison to my other machines.

Perhaps I could change the suggested edit to mention it should be done if the user plans on using IWD in their final install?

flokli:
I also realized we include nixpkgs' `installation-device.nix`, so would have NetworkManager enabled.

We however then do some patching in  `iso-configuration/installer-configuration.nix` to disable it again, instead of setting `networking.networkmanager.wifi.backend = "iwd";`, which would probably be the correct way.

So I think `iso-configuration/installer-configuration.nix` should be changed to set that, which should automatically remove the need for setting most of the other wireless-related options (including `services.getty.helpLine`), and get things to work without surprises.

We then could suggest the same networkmanager config in the docs.

WillemToorenburgh:
That makes good sense! Do you mind if I take a stab at implementing that?

I do still think it's important to also correctly set IWD's DNS backend, unless the installer environment uses systemd-resolved. Perhaps I could drop that in there too while I'm at it?

flokli:
> That makes good sense! Do you mind if I take a stab at implementing that?

Gladly!

> I do still think it's important to also correctly set IWD's DNS backend, unless the installer environment uses systemd-resolved. Perhaps I could drop that in there too while I'm at it?

Let's try it out - I'm running a pretty vanilla networkmanager config here, where I only configured it to use the iwd backend. I sometimes run resolved, but I haven't noticed it not working if I configure custom DNS servers.

Maybe "networking.wireless.iwd.settings.General.EnableNetworkConfiguration" is the culprit here, and somehow interferes with NetworkManagers' job? Do you have that set?

WillemToorenburgh:
Ooooooh. I do indeed have `networking.wireless.iwd.settings.General.EnableNetworkConfiguration` set to `true`. I don't know if that explains why I wasn't able to set DNS servers in Plasma Settings... my first thought was that IWD's and NetworkManager's DHCP clients would conflict, but surely the folks working on NetworkManager's IWD backend would have thought of that scenario and made the setup use IWD's client for wireless networks?

WillemToorenburgh:
@flokli I've committed two revisions: clarified the docs edit to say it's for final installations, and implemented the guidance in the installer environment!

iso-configuration/installer-configuration.nix
flokli: I don't think we still need to set this

iso-configuration/installer-configuration.nix
```diff
@@ -91,8 +91,11 @@
   networking.wireless.iwd = {
     enable = true;
     settings.General.EnableNetworkConfiguration = true;
```
flokli: … and it might be less error-prone if we don't set this either, not sure about how NetworkManager deals with this being set.

iso-configuration/installer-configuration.nix
```diff
@@ -91,8 +91,11 @@
...
+    # NixOS by default uses resolvconf for DNS, but iwd's default is systemd-resolved. Make iwd use resolvconf
+    settings.Network.NameResolvingService = "resolvconf";
   };
-  networking.networkmanager.enable = lib.mkForce false;
+  # Make sure NetworkManager, also enabled by default, uses iwd for wifi
```
flokli: The comment from Line 88 can probably be moved here, if we delete the rest of the options we set.

WillemToorenburgh:
I've applied your recommendations in the installer code, and updated the documentation to match. Always nice when improved functionality lets us remove lines overall :D

flokli (approved):
LGTM (we can always improve on the resolvconf situation), but I'd like to get a second opinion.

WillemToorenburgh:
@flokli a small update for you! I've been running with   `networking.wireless.iwd.settings.General.EnableNetworkConfiguration` unset for a while now, including joining a new network via Plasma Settings, and everything has been working great. I think we're safe to say that it should indeed be left unset (disabled) by default.

Also, @yuyuyureka, sorry to ping you directly (hello, nice to meet you!), but I saw this comment of yours in the now locked https://github.com/nix-community/nixos-apple-silicon/issues/414#issuecomment-4113022499, where you say
> [...] please feel free to open a PR to re-enable NetworkManager support

Is this PR accomplishing that? Or is NetworkManager additionally disabled elsewhere in the repo?

iso-configuration/installer-configuration.nix
WillemToorenburgh: Updated here: https://github.com/nix-community/nixos-apple-silicon/pull/433#issuecomment-4232608360

Resolving thread as it seems leaving this value unset is indeed the correct thing to do.

yuyuyureka:
Hello! :)
Afaik NetworkManager was only explicitly disabled in the installer-configuration.nix. So this would indeed achieve what I meant and make it possible to configure the WiFi (or Wired) connection using nmcli like it's nowadays the default in the upstream NixOS minimal image as well.

flokli:
@WillemToorenburgh so we might not even need to set `networking.wireless.iwd.settings.Network.NameResolvingService` if `EnableNetworkConfiguration` is not set?

In that case, can we simplify things further here?

WillemToorenburgh:
That's a very good point. I'll test this later today and see if anything changes

WillemToorenburgh:
Things seem to work! Good thinking, that setting must be part of the network configuration component. I'll revise the PR further.

WillemToorenburgh:
Done! The PR is now largely deletions 😅

WillemToorenburgh:
Updated the title and description to better match what's actually happening. Thanks again for iterating on this with me 😄

flokli:
> Perfection is achieved, not when there is nothing more to add, but when there is nothing left to take away.

(Antoine de Saint-Exupéry, Airman's Odyssey)

PR #434, opened 2026-03-15
Change default option for Asahi Linux support to false

eureka-cpu:
This change would make it so that users with multi-machine configurations don't need to guard against this module, only to re-enable the option later for the machine that actually needs it, eg.

```nix
# flake.nix
nixosModules.asahi = { config, lib, ... }: {
  imports = [ nixos-apple-silicon.nixosModules.default ];
  config.nixpkgs.overlays = [ nixos-apple-silicon.overlays.default ];
  config.hardware.asahi.enable = lib.mkDefault false;
};
```
```nix
# configuration.nix
hardware.asahi.enable = true;
```

The only change current users will then need to make would be to enable the module for their machine.

### DIFF
--- a/apple-silicon-support/modules/default.nix
+++ b/apple-silicon-support/modules/default.nix
@@ -45,7 +45,7 @@
   options.hardware.asahi = {
     enable = lib.mkOption {
       type = lib.types.bool;
-      default = true;
+      default = false;
       description = ''
         Enable the basic Asahi Linux components, such as kernel and boot setup.
       '';


flokli:
Mmmh, I sympathize, but this might break lots of systems who don't set this to true (pretty much all of them, as it got introduced later on and the installation guide also doesn't advice to explicitly set it to true:

```
This release also introduces a `hardware.asahi.enable` configuration option,
which defaults to true. Setting this option to false disables all effects of
the Apple Silicon support module (including ignoring all other options), which
may be useful for multi-system configurations.
```

Is it possible to show an eval warning if the user didn't set it explicitly?

In that case, we could show a warning that this will default to false at some point, encouraging users to manually set it to true (as well as updating all installation documentation and mentioning this in the upcoming release notes).

And one release later, we can actually flip the default.

eureka-cpu:
Yeah that makes sense. I don't have the time at the moment, but I've marked this so that maintainers can edit the pull request, if you'd like to add the warning you're of course welcome to. If not, I can update it once I have the time. I just used the github edit feature to quickly fork and open this so that a maintainer would be aware of the issue.

flokli:
https://github.com/nix-community/nixos-apple-silicon/pull/499

PR #444, opened 2026-04-10
Add Dependabot for nix flake updates

JamieMagee:
See https://github.blog/changelog/2026-04-07-dependabot-version-updates-now-support-the-nix-ecosystem/


zvolin:
this sounds nice, I like keeping my nixpkgs at the same commit as apple-silicon to ensure everything is smooth, and taken from cache, but that locks my updates to the flake.lock updates

PR #451, opened 2026-04-28
Support Asahi installer vendorfw format

Solidsilver:
Asahi installer moved peripheral firmware from `/boot/asahi` to `/boot/vendorfw` and changed the packaging. I think this was changed in [this commit](https://github.com/AsahiLinux/asahi-installer/commit/ee574521ea00601735c064fd87ca549042734728), but as mentioned in the latest [asahi blog](https://asahilinux.org/2026/04/progress-report-7-0/) the deployed installer hasn't been updated in a while. When running the "Rebuild vendor firmware package" option on the asahi installer, I got this new package format and ran into issues trying to apply the firmware changes.

This branch updates both the runtime firmware module and the installer ISO to handle the new format.

Changes:

• peripheralFirmwareDirectory default now searches /boot/vendorfw first, then falls back to /boot/asahi and the equivalent installer mount paths
• Build scripts and installer postBootCommands auto-detect which format is present and extract accordingly. (Note: I have not tested the installer yet.)

Tested on M1 Pro by pointing at this branch and re-building. I have not tested the installer changes.

Disclaimer: `Kimi K2.6` was used in helping to diagnose the issue and suggesting script changes. I manually reviewed, tested, and made a few teaks to make the code a bit more readable.

### DIFF
--- a/apple-silicon-support/modules/peripheral-firmware/default.nix
+++ b/apple-silicon-support/modules/peripheral-firmware/default.nix
@@ -4,78 +4,241 @@
   lib,
   ...
 }:
-{
-  config = lib.mkIf config.hardware.asahi.enable {
-    assertions = lib.mkIf config.hardware.asahi.extractPeripheralFirmware [
+
+let
+  cfg = config.hardware.asahi;
+
+  # Script to extract Asahi firmware
+  # during boot
+  extractAsahiFirmwareInitrdScript = pkgs.writeShellScript "asahi-firmware-extract-initrd" ''
+    set -u
+    mkdir -p /run/asahi-firmware
+
+    # m1n1 may stage vendor firmware directly into the initramfs as /vendorfw.
+    # When that happens, skip ESP discovery and use it directly.
+    if [ -d /vendorfw ] && [ -n "$(ls -A /vendorfw 2>/dev/null)" ]; then
+      echo "Bootloader-staged vendor firmware found in /vendorfw"
+      cp -a /vendorfw/. /run/asahi-firmware/
+      if [ -f /sys/module/firmware_class/parameters/path ]; then
+        echo -n "/run/asahi-firmware" > /sys/module/firmware_class/parameters/path
+        echo "Registered /run/asahi-firmware as kernel firmware search path"
+      fi
+      exit 0
+    fi
+
+    esp_partuuid=""
+    read -r -d ''' esp_partuuid < /proc/device-tree/chosen/asahi,efi-system-partition 2>/dev/null || true
+    if [ -z "$esp_partuuid" ]; then
+      echo "Warning: could not determine Asahi ESP partition UUID"
+      exit 0
+    fi
+
+    esp_dev="/dev/disk/by-partuuid/$esp_partuuid"
+
+    # Wait up to 3 seconds for udev to create the block device symlink
+    i=0
+    while [ $i -lt 30 ]; do
+      if [ -e "$esp_dev" ]; then
+        break
+      fi
+      sleep 0.1
+      i=$((i + 1))
+    done
+    if [ ! -e "$esp_dev" ]; then
+      echo "Warning: ESP block device $esp_dev not found, skipping firmware extraction"
+      exit 0
+    fi
+
+    mkdir -p /tmp/asahi-esp
+    if ! mount -o ro -t vfat "$esp_dev" /tmp/asahi-esp 2>/dev/null; then
+      # Fallback: ESP may already be mounted (e.g. custom initrd config).
+      # Resolve the real backing device and look it up in /proc/mounts.
+      real_dev=""
+      real_dev=$(readlink -f "$esp_dev" 2>/dev/null || true)
+      existing=""
+      if [ -n "$real_dev" ] && [ -r /proc/mounts ]; then
+        while read -r dev mnt _; do
+          if [ "$dev" = "$esp_dev" ] || [ "$dev" = "$real_dev" ]; then
+            existing="$mnt"
+            break
+          fi
+        done < /proc/mounts
+      fi
+      if [ -n "$existing" ] && mount --bind "$existing" /tmp/asahi-esp; then
+        echo "ESP already mounted at $existing, bind-mounting for firmware extraction"
+      else
+        echo "Warning: failed to mount ESP, skipping firmware extraction"
+        rmdir /tmp/asahi-esp 2>/dev/null || true
+        exit 0
+      fi
+    fi
+
+    cleanup() {
+      umount /tmp/asahi-esp 2>/dev/null || true
+      rmdir /tmp/asahi-esp 2>/dev/null || true
+    }
+    trap cleanup EXIT
+
+    echo "Extracting Asahi firmware from ESP..."
+
+    if [ -f /tmp/asahi-esp/vendorfw/firmware.cpio ]; then
+      mkdir -p /tmp/asahi-fwextract
+      (
+        cd /tmp/asahi-fwextract
+        ${pkgs.cpio}/bin/cpio -id --quiet --no-absolute-filenames < /tmp/asahi-esp/vendorfw/firmware.cpio
+      )
+      if [ -d /tmp/asahi-fwextract/vendorfw ]; then
+        cp -a /tmp/asahi-fwextract/vendorfw/. /run/asahi-firmware/
+        echo "Asahi firmware extracted successfully"
+        if [ -f /sys/module/firmware_class/parameters/path ]; then
+          echo -n "/run/asahi-firmware" > /sys/module/firmware_class/parameters/path
+          echo "Registered /run/asahi-firmware as kernel firmware search path"
+        fi
+      else
+        echo "Warning: firmware archive did not contain vendorfw/ directory"
+      fi
+    elif [ -f /tmp/asahi-esp/asahi/all_firmware.tar.gz ]; then
+      echo "Warning: legacy all_firmware.tar.gz found on ESP but boot-time extraction only supports vendorfw/firmware.cpio"
+    else
+      echo "Warning: vendorfw/firmware.cpio not found on ESP"
+    fi
+    exit 0
+  '';
+
+  # Script to extract Asahi firmware
+  # at eval time (not boot-time)
+  extractAsahiFirmwareAtEval =
+    let
+      pkgs' = cfg.pkgs;
+      firmwareDir = cfg.peripheralFirmwareDirectory;
+    in
+    pkgs.runCommand "asahi-peripheral-firmware"
       {
-        assertion = config.hardware.asahi.peripheralFirmwareDirectory != null;
-        message = ''
-          Asahi peripheral firmware extraction is enabled but the firmware
-          location appears incorrect.
-        '';
-      }
-    ];
-
-    hardware.firmware =
-      let
-        pkgs' = config.hardware.asahi.pkgs;
-      in
-      lib.mkIf
-        (
-          (config.hardware.asahi.peripheralFirmwareDirectory != null)
-          && config.hardware.asahi.extractPeripheralFirmware
-        )
-        [
-          (pkgs.stdenv.mkDerivation {
-            name = "asahi-peripheral-firmware";
-
-            nativeBuildInputs = [
-              pkgs'.asahi-fwextract
-              pkgs.cpio
-            ];
-
-            buildCommand = ''
-              mkdir extracted
-              asahi-fwextract ${config.hardware.asahi.peripheralFirmwareDirectory} extracted
-
-              mkdir -p $out/lib/firmware
-              cat extracted/firmware.cpio | cpio -id --quiet --no-absolute-filenames
-              mv vendorfw/* $out/lib/firmware
-            '';
-          })
+        nativeBuildInputs = [
+          pkgs'.asahi-fwextract
+          pkgs.cpio
         ];
-  };
+      }
+      ''
+        mkdir -p $out/lib/firmware
+
+        if [ -f "${firmwareDir}/firmware.cpio" ]; then
+          cpio_src="${firmwareDir}/firmware.cpio"
+
+        elif [ -f "${firmwareDir}/all_firmware.tar.gz" ]; then
+          mkdir extracted
+          asahi-fwextract "${firmwareDir}" extracted
+          cpio_src=extracted/firmware.cpio
+
+        else
+          echo "ERROR: No recognized Asahi firmware format found in ${firmwareDir}" >&2
+          echo "Expected: firmware.cpio (installer 0.8.0+) or all_firmware.tar.gz (legacy)" >&2
+          exit 1
+        fi
+
+        cpio -id --quiet --no-absolute-filenames < "$cpio_src"
+        cp -a vendorfw/. $out/lib/firmware
+      '';
+in
+{
+  config = lib.mkIf cfg.enable (
+    lib.mkMerge [
+      {
+        assertions = [
+          {
+            assertion = !cfg.extractPeripheralFirmware || cfg.peripheralFirmwareDirectory != null;
+            message = ''
+              Asahi peripheral firmware extraction is enabled but the firmware
+              location appears incorrect.
+            '';
+          }
+        ];
+      }
+
+      (lib.mkIf cfg.extractPeripheralFirmware {
+        hardware.firmware = lib.mkIf (cfg.peripheralFirmwareDirectory != null) [
+          extractAsahiFirmwareAtEval
+        ];
+      })
+
+      (lib.mkIf (!cfg.extractPeripheralFirmware) {
+        # vfat + codepage modules for ESP mounting in initrd
+        boot.initrd.availableKernelModules = [
+          "vfat"
+          "nls_cp437"
+          "nls_iso8859-1"
+        ];
+
+        # Stage 1 systemd service: mount ESP, extract vendorfw, register firmware path.
+        # The script and cpio binary must be present in the initrd.
+        boot.initrd.systemd.storePaths = [
+          pkgs.cpio
+          extractAsahiFirmwareInitrdScript
+        ];
+
+        boot.initrd.systemd.services.asahi-firmware-extract = {
+          description = "Extract Asahi peripheral firmware from ESP";
+          wantedBy = [ "initrd.target" ];
+          after = [ "systemd-udevd.service" ];
+          before = [
+            "systemd-modules-load.service"
+            "initrd-switch-root.target"
+          ];
+          unitConfig.DefaultDependencies = false;
+          serviceConfig = {
+            Type = "oneshot";
+            RemainAfterExit = true;
+            ExecStart = extractAsahiFirmwareInitrdScript;
+          };
+        };
+
+        # NixOS's udev activation snippet (nixpkgs:
+        # nixos/modules/services/hardware/udev.nix, the firmware-loading-path
+        # block) unconditionally overwrites
+        # /sys/module/firmware_class/parameters/path with the Nix-store combined
+        # firmware directory. Re-point it at the initrd-extracted firmware so
+        # Wi-Fi/Bluetooth drivers find their blobs. Must run after "udevd" but
+        # before systemd-modules-load triggers driver binds.
+        system.activationScripts.asahi-firmware-path = lib.stringAfter [ "udevd" ] ''
+          if [ -d /run/asahi-firmware ] && [ -e /sys/module/firmware_class/parameters/path ]; then
+            echo -n "/run/asahi-firmware" > /sys/module/firmware_class/parameters/path
+          fi
+        '';
+      })
+    ]
+  );
 
   options.hardware.asahi = {
     extractPeripheralFirmware = lib.mkOption {
       type = lib.types.bool;
-      default = true;
+      default = false;
       description = ''
-        Automatically extract the non-free non-redistributable peripheral
-        firmware necessary for features like Wi-Fi.
+        When enabled, extract the non-free non-redistributable
+        Asahi peripheral firmware into the Nix store at evaluation time.
+
+        Enable this if you are using the legacy <filename>asahi/all_firmware.tar.gz</filename>
+        format or want declarative/offline firmware management.
+
+        The default (disabled) loads firmware from the ESP at boot time, which
+        is the recommended approach and matches upstream Fedora Asahi Linux.
       '';
     };
 
     peripheralFirmwareDirectory = lib.mkOption {
       type = lib.types.nullOr lib.types.path;
 
-      default = lib.findFirst (path: builtins.pathExists (path + "/all_firmware.tar.gz")) null [
-        # path when the system is operating normally
-        /boot/asahi
-        # path when the system is mounted in the installer
-        /mnt/boot/asahi
-      ];
+      default = null;
 
       description = ''
         Path to the directory containing the non-free non-redistributable
-        peripheral firmware necessary for features like Wi-Fi. Ordinarily, this
-        will automatically point to the appropriate location on the ESP. Flake
-        users and those interested in maximum purity will want to copy those
-        files elsewhere and specify this manually.
+        peripheral firmware necessary for features like Wi-Fi.
 
-        Currently, this consists of the files `all-firmware.tar.gz` and
-        `kernelcache*`. The official Asahi Linux installer places these files
-        in the `asahi` directory of the EFI system partition when creating it.
+        This option is only used when <option>extractPeripheralFirmware</option>
+        is enabled. When boot-time loading is used (the default), firmware is read
+        directly from the ESP and this option is ignored.
+
+        For declarative management, copy the firmware files elsewhere (like ./firmware) and
+        specify this path manually.
       '';
     };
   };
--- a/apple-silicon-support/modules/peripheral-firmware/default.nix
+++ b/apple-silicon-support/modules/peripheral-firmware/default.nix
@@ -152,6 +152,9 @@ in
           "nls_iso8859-1"
         ];
 
+        boot.initrd.kernelModules = [ "overlay" ];
+        boot.kernelModules = [ "overlay" ];
+
         # Stage 1 systemd service: mount ESP, extract vendorfw, register firmware path.
         # The script and cpio binary must be present in the initrd.
         boot.initrd.systemd.storePaths = [
@@ -175,16 +178,25 @@ in
           };
         };
 
-        # NixOS's udev activation snippet (nixpkgs:
-        # nixos/modules/services/hardware/udev.nix, the firmware-loading-path
-        # block) unconditionally overwrites
-        # /sys/module/firmware_class/parameters/path with the Nix-store combined
-        # firmware directory. Re-point it at the initrd-extracted firmware so
-        # Wi-Fi/Bluetooth drivers find their blobs. Must run after "udevd" but
-        # before systemd-modules-load triggers driver binds.
-        system.activationScripts.asahi-firmware-path = lib.stringAfter [ "udevd" ] ''
+        # udev activation overwrites the firmware path with the NixOS combined
+        # directory, hiding the Asahi firmware extracted in initrd. Overlay
+        # both so user-declared firmware (hardware.firmware) remains available.
+        # Must run after udevd but before module loading.
+        system.activationScripts.asahi-firmware-path = lib.stringAfter [ "udevd" "modprobe" ] ''
           if [ -d /run/asahi-firmware ] && [ -e /sys/module/firmware_class/parameters/path ]; then
-            echo -n "/run/asahi-firmware" > /sys/module/firmware_class/parameters/path
+            nixos_fw=$(cat /sys/module/firmware_class/parameters/path)
+
+            umount -l /run/firmware-merged 2>/dev/null || true
+            mkdir -p /run/firmware-merged
+
+            # Asahi firmware takes precedence if both sources provide the same path.
+            if [ -n "$nixos_fw" ] && [ -d "$nixos_fw" ] && mount -t overlay overlay -o ro,lowerdir=/run/asahi-firmware:"$nixos_fw" /run/firmware-merged; then
+              echo -n "/run/firmware-merged" > /sys/module/firmware_class/parameters/path
+            else
+              # If overlay fails, keep built-in Wi-Fi/BT working with Asahi firmware.
+              echo "Warning: failed to merge Asahi and NixOS firmware; falling back to Asahi firmware only" >&2
+              echo -n "/run/asahi-firmware" > /sys/module/firmware_class/parameters/path
+            fi
           fi
         '';
       })
--- a/apple-silicon-support/modules/peripheral-firmware/default.nix
+++ b/apple-silicon-support/modules/peripheral-firmware/default.nix
@@ -8,23 +8,10 @@
 let
   cfg = config.hardware.asahi;
 
-  # Script to extract Asahi firmware
-  # during boot
+  # Script to extract Asahi firmware during boot
   extractAsahiFirmwareInitrdScript = pkgs.writeShellScript "asahi-firmware-extract-initrd" ''
     set -u
     mkdir -p /run/asahi-firmware
-
-    # m1n1 may stage vendor firmware directly into the initramfs as /vendorfw.
-    # When that happens, skip ESP discovery and use it directly.
-    if [ -d /vendorfw ] && [ -n "$(ls -A /vendorfw 2>/dev/null)" ]; then
-      echo "Bootloader-staged vendor firmware found in /vendorfw"
-      cp -a /vendorfw/. /run/asahi-firmware/
-      if [ -f /sys/module/firmware_class/parameters/path ]; then
-        echo -n "/run/asahi-firmware" > /sys/module/firmware_class/parameters/path
-        echo "Registered /run/asahi-firmware as kernel firmware search path"
-      fi
-      exit 0
-    fi
 
     esp_partuuid=""
     read -r -d ''' esp_partuuid < /proc/device-tree/chosen/asahi,efi-system-partition 2>/dev/null || true
@@ -109,13 +96,12 @@ let
   # at eval time (not boot-time)
   extractAsahiFirmwareAtEval =
     let
-      pkgs' = cfg.pkgs;
       firmwareDir = cfg.peripheralFirmwareDirectory;
     in
     pkgs.runCommand "asahi-peripheral-firmware"
       {
         nativeBuildInputs = [
-          pkgs'.asahi-fwextract
+          cfg.pkgs.asahi-fwextract
           pkgs.cpio
         ];
       }
@@ -141,27 +127,24 @@ let
       '';
 in
 {
+  imports = [
+    (lib.mkRemovedOptionModule [ "hardware" "asahi" "extractPeripheralFirmware" ] ''
+      Eval-time firmware extraction is now enabled by setting
+      `hardware.asahi.peripheralFirmwareDirectory` to a path.
+      Leave it as `null` (the default) to load firmware from
+      the ESP at boot time (recommended).
+    '')
+  ];
+
   config = lib.mkIf cfg.enable (
     lib.mkMerge [
       {
-        assertions = [
-          {
-            assertion = !cfg.extractPeripheralFirmware || cfg.peripheralFirmwareDirectory != null;
-            message = ''
-              Asahi peripheral firmware extraction is enabled but the firmware
-              location appears incorrect.
-            '';
-          }
-        ];
-      }
-
-      (lib.mkIf cfg.extractPeripheralFirmware {
         hardware.firmware = lib.mkIf (cfg.peripheralFirmwareDirectory != null) [
           extractAsahiFirmwareAtEval
         ];
-      })
+      }
 
-      (lib.mkIf (!cfg.extractPeripheralFirmware) {
+      (lib.mkIf (cfg.peripheralFirmwareDirectory == null) {
         # vfat + codepage modules for ESP mounting in initrd
         boot.initrd.availableKernelModules = [
           "vfat"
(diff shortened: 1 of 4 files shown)


autrimpo:
I tried a similar change yesterday, just simpler (always using the `firmware.cpio` from vendorfw) and it seems to work fine.

For ALS, you need to enable it in the kernel, i.e.:
```
boot.kernelPatches = [
  {
    name = "Asahi ALS config";
    patch = null;
    structuredExtraConfig = with lib.kernel; {
      RUST_FW_LOADER_ABSTRACTIONS = yes;
      IIO_AOP_SENSOR_ALS = module;
    };
  }
];
```

flokli:
This needing to be enabled explicitly is probably not intentional, see [IRC](https://oftc.catirclogs.org/asahi/2026-04-29#1777492503-1777493365;) and https://github.com/AsahiLinux/linux/pull/490.

Solidsilver:
Ah, good to know! After applying the kernel patches from @autrimpo, ALS did start working for me, so hopefully that upstream fix gets merged soon. 

Just to keep the scope clear: The ALS kernel issue is tangential to this PR, which strictly focuses on handling the new Asahi firmware format changes.

Regarding the firmware changes—@autrimpo, you mentioned a simpler approach of strictly using vendorfw. I included the fallback to /boot/asahi to maintain backward compatibility with older installer formats just in case, but if the maintainers prefer a hard cutover to the new directory, I'm happy to strip the fallback logic out to simplify.

flokli:
> Just to keep the scope clear: The ALS kernel issue is tangential to this PR, which strictly focuses on handling the new Asahi firmware format changes.

Understood, though doing the firmware dance again to be able to use it is a big argument for thinking about how to manage firmware bits.

I personally prefer having the asahi installer take care of managing the firmware bits, but I'm also worried about the migration path. I'd like to think about it more.

> Regarding the firmware changes—@autrimpo, you mentioned a simpler approach of strictly using vendorfw. I included the fallback to /boot/asahi to maintain backward compatibility with older installer formats just in case, but if the maintainers prefer a hard cutover to the new directory, I'm happy to strip the fallback logic out to simplify.

Can you push that approach to a second branch, so I can look at it too?

Solidsilver:
@flokli I've pushed that second approach without fallback: https://github.com/Solidsilver/nixos-apple-silicon/tree/feat/vendorfw-support-nofallback

I believe this will force users to run a firmware extract from the latest Asahi installer before being able to switch.

Note: I have not tested the install process yet, just a switch from a live system. My changes to the install script are based on what I think should happen, so some extra eyes there would be appreciated.

flokli:
> @flokli I've pushed that second approach without fallback: https://github.com/Solidsilver/nixos-apple-silicon/tree/feat/vendorfw-support-nofallback
> 
> I believe this will force users to run a firmware extract from the latest Asahi installer before being able to switch.

This reads from the firmware directory at **eval time**.

So users would need to re-eval their system on their macbooks after running the installer to be able to actually get the firmware. I don't think this makes much sense.

[Compare with Fedora Asahi Linux](https://asahilinux.org/2026/04/progress-report-7-0/):

> To achieve this, the Asahi Installer gathers up all the firmware it knows we will need in Linux and stores it on the EFI System Partition it creates. A Dracut module then mounts this to a subdirectory of /lib/firmware/, where drivers can find it.

If we consider the firmware on the ESP to be just firmware that's managed externally, it shouldn't need to be piped back into eval. We'd need to check our initrd logic and make sure the firmware loading machinery can discover firmware extracted by the asahi installer into the ESP.

We can probably use the NixOS option for backwards-compat until users switched to that approach, or for users that want to manage the firmware declaratively using Nix, but the new code reading the firmware from the ESP should happen at boot, not eval time.

autrimpo:
> Just to keep the scope clear: The ALS kernel issue is tangential to this PR, which strictly focuses on handling the new Asahi firmware format changes.

Of course, I just mentioned it so it's easier to test the firmware.

> Regarding the firmware changes—@autrimpo, you mentioned a simpler approach of strictly using vendorfw.

I was just trying to get the ALS working quickly, so I copied the cpio file to my flake and did the simplest thing I could think of:

```
hardware.asahi.extractPeripheralFirmware = false;
hardware.firmware = [
  (pkgs.stdenv.mkDerivation {
    name = "asahi-peripheral-firmware";

    nativeBuildInputs = [
      pkgs.cpio
    ];

    buildCommand = ''
      mkdir -p $out/lib/firmware
      cat ${./firmware.cpio} | cpio -id --quiet --no-absolute-filenames
      mv vendorfw/* $out/lib/firmware
    '';
  })
];
```

This is definitely not the way to do it for nixos-asahi, for reasons @flokli just mentioned.

Solidsilver:
> We can probably use the NixOS option for backwards-compat until users switched to that approach, or for users that want to manage the firmware declaratively using Nix, but the new code reading the firmware from the ESP should happen at boot, not eval time.

I agree, I think that'd reduce friction for users who don't want declarative firmware management. Would you want to merge in these changes for now as an iterative step that at least gets the `vendorfw` format working with backwards compat? Then the boot handling could be added later. Or do you want to get all of the changes in at once? I might be able to take a stab at that sometime in the next week.

flokli:
I'd rather not introduce the new vendorfw path to be collected at eval time, but only use it for the new boot-time approach. It feels much cleaner and predictable to users.

waltmck:
@Solidsilver Alternatively, if you are up for it, now would be a great opportunity to switch to the approach discussed [here](https://github.com/nix-community/nixos-apple-silicon/pull/94#issuecomment-4121267728). It should be possible to write a FOD that downloads the firmware from Apple's servers with range queries at build time, using @GavBog's [AEA tools](https://github.com/GavBog/aea-tools).

flokli:
> @Solidsilver Alternatively, if you are up for it, now would be a great opportunity to switch to the approach discussed [here](https://github.com/nix-community/nixos-apple-silicon/pull/94#issuecomment-4121267728). It should be possible to write a FOD that downloads the firmware from Apple's servers with range queries at build time, using @GavBog's [AEA tools](https://github.com/GavBog/aea-tools).

IIRC we also need per-device calibrations, and these can't be downloaded from apples servers.

GavBog:
> > @Solidsilver Alternatively, if you are up for it, now would be a great opportunity to switch to the approach discussed [here](https://github.com/nix-community/nixos-apple-silicon/pull/94#issuecomment-4121267728). It should be possible to write a FOD that downloads the firmware from Apple's servers with range queries at build time, using @GavBog's [AEA tools](https://github.com/GavBog/aea-tools).
> 
> IIRC we also need per-device calibrations, and these can't be downloaded from apples servers.

We're only pulling the Wifi, Bluetooth, and Input firmware (and appleh13camerad). The IPSW has all the firmware we need for all macbook models so, as far as I know, there should be no issue using it directly.

My proof of concept (in the examples directory of aea-tools) works well on my Macbook M1 Air, though it is not ready to be used here since I had to remove some Wifi Firmware used by the M2 lineup (some files from firmware/wifi/C-4388__s-C0) since it was erroring with asahi-fwextract.
I don't have an M2 mac to test on but assuming it doesn't just work the first step to debugging would be removing the `is_incompatible_wifi_item()` function from the example.

It should also be noted that aea-tools is built for extracting from the AEA (Apple Encrypted Archive) format used to wrap the RootFS in macOS 15+ and will not work with prior versions that don't use this wrapping. The asahi team tests with version 13.5 so it is diverging from that.

flokli:
> > > @Solidsilver Alternatively, if you are up for it, now would be a great opportunity to switch to the approach discussed [here](https://github.com/nix-community/nixos-apple-silicon/pull/94#issuecomment-4121267728). It should be possible to write a FOD that downloads the firmware from Apple's servers with range queries at build time, using @GavBog's [AEA tools](https://github.com/GavBog/aea-tools).
> > 
> > 
> > IIRC we also need per-device calibrations, and these can't be downloaded from apples servers.
> 
> We're only pulling the Wifi, Bluetooth, and Input firmware (and appleh13camerad). The IPSW has all the firmware we need for all macbook models so, as far as I know, there should be no issue using it directly.

Nop, `iio_aop_als` needs a calibration file:

```
[   13.775020] iio_aop_als als.1.auto: Direct firmware load for apple/aop-als-cal.bin failed with error -2
```

Solidsilver:
@flokli I've made the changes necessary for boottime extraction of firmware. I took a look at how it's done on Fedora and tried to translate that here somewhat. Right now it only loads the new `vendorfw` firmware format at boottime, and declarative extraction at build-time works for the old and new formats.

I did have to implement a "hack" of sorts to get the extracted firmware picked up. The udev script overwrites /sys/module/firmware_class/parameters/path with the path to the Nix-store firmware, so I had to write a script to run after udev to re-point at the extracted asahi firmware.

I tested both boottime extraction and declarative build-time extraction. From what I can tell everything works. Would definitely appreciate eyes and testers.

eva-val:
M2 air 15 in macos 26.4.1
Booted into macos to do the asahi fw dump thing from the blog post
>Invoke the Asahi Installer by running curl https://alx.sh | sh
Choose the “Rebuild vendor firmware package” option when prompted

Rebooted into nixos. Pinned my flake to this PR and rebuilt/rebooted.

Appears to have picked it up
```
❯ ls -la /boot/vendorfw/
total 63752
drwxr-xr-x  3 root root     4096 Apr 30 12:22 .
drwxr-xr-x 10 root root     4096 Dec 31  1969 ..
-rwxr-xr-x  1 root root     4096 Feb  2 12:45 ._firmware.cpio
-rwxr-xr-x  1 root root 32531640 Apr 30 12:22 firmware.cpio
-rwxr-xr-x  1 root root     4096 Feb  2 12:45 ._firmware.tar
-rwxr-xr-x  1 root root 32675840 Apr 30 12:22 firmware.tar
-rwxr-xr-x  1 root root     4096 Feb  2 12:45 ._manifest.txt
-rwxr-xr-x  1 root root    38551 Apr 30 12:22 manifest.txt
-rwxr-xr-x  1 root root     4096 Feb  2 12:45 ._u-boot
drwxr-xr-x  3 root root     4096 Apr 30 12:22 u-boot
```

```
❯ journalctl -b 0 -u asahi-firmware-extract.service --no-pager


Mar 17 12:55:52 nixbook systemd[1]: Starting Extract Asahi peripheral firmware from ESP...
May 01 12:40:00 nixbook 2r49v9f2gwahg4sqgkh6vg4mh32bd1z9-asahi-firmware-extract-initrd[207]: Extracting Asahi firmware from ESP...
May 01 12:40:00 nixbook 2r49v9f2gwahg4sqgkh6vg4mh32bd1z9-asahi-firmware-extract-initrd[207]: Asahi firmware extracted successfully
May 01 12:40:00 nixbook 2r49v9f2gwahg4sqgkh6vg4mh32bd1z9-asahi-firmware-extract-initrd[207]: Registered /run/asahi-firmware as kernel firmware search path
May 01 12:40:00 nixbook systemd[1]: Finished Extract Asahi peripheral firmware from ESP.
May 01 12:40:01 nixbook systemd[1]: asahi-firmware-extract.service: Deactivated successfully.
May 01 12:40:01 nixbook systemd[1]: Stopped Extract Asahi peripheral firmware from ESP.
May 01 12:40:01 nixbook systemd[1]: asahi-firmware-extract.service: Consumed 76ms CPU time over 1.236s wall clock time, 101.9M memory peak.
```
```
❯ journalctl -b 0 -k --grep='apple_aop' --no-pager
May 01 12:40:03 nixbook kernel: apple_aop 24ac00000.aop: RTKit: Initializing (protocol version 12)
May 01 12:40:03 nixbook kernel: apple_aop 24ac00000.aop: RTKit: syslog message: AOP_oslog_init.cpp:96: Start AOP OSLog init
May 01 12:40:03 nixbook kernel: apple_aop 24ac00000.aop: RTKit: syslog message: P_os_log_support.cpp:60: AOP os log initialized
May 01 12:40:03 nixbook kernel: apple_aop 24ac00000.aop: RTKit: syslog message: SPU_dram_client.cpp:211: AOP platform dram client initialized
May 01 12:40:03 nixbook kernel: apple_aop 24ac00000.aop: RTKit: syslog message: ALSCT720.cpp:57: probe
May 01 12:40:03 nixbook kernel: apple_aop 24ac00000.aop: RTKit: syslog message: ALSCT720.cpp:72: UNSUPPORTED HW
May 01 12:40:03 nixbook kernel: apple_aop 24ac00000.aop: RTKit: syslog message: ALSCT720.cpp:85: CT720 device NOT found.
May 01 12:40:03 nixbook kernel: apple_aop 24ac00000.aop: RTKit: syslog message: ALSVD6286.cpp:80: [ALS2] VD6286 device found. Calibration Override:0
May 01 12:40:03 nixbook kernel: apple_aop 24ac00000.aop: RTKit: syslog message: iggerController.cpp:129: VoiceTrig ready!! <AppleAOPVoiceTriggerController v240.2 built: Jun 30 202317:4
May 01 12:40:03 nixbook kernel: apple_aop 24ac00000.aop: RTKit: syslog message: SPU_init.cpp:320: ===( AOP IS READY )===
```
```
❯ cat /sys/bus/iio/devices/iio:device1/in_illuminance_input

44
```

Now to see if i can get auto brightness working..

docs/uefi-standalone.md
```diff
@@ -243,14 +243,15 @@ Add the `./apple-silicon-support` directory to the imports list and switch off t
...
 
-Currently, the only supported way to update the peripheral firmware files is to destroy and re-create the EFI system partition, so they will not change unexpectedly. If you do not want the impurity of referencing them (or are using flakes where this is prohibited), copy them off the EFI system partition (e.g. on the installation ISO `mkdir -p /mnt/etc/nixos/firmware && cp /mnt/boot/asahi/{all_firmware.tar.gz,kernelcache*} /mnt/etc/nixos/firmware`) and specify this path in your configuration:
+By default, the Apple Silicon support module now loads peripheral firmware from the ESP at boot time (before any drivers that need it are loaded). This matches the approach used by Fedora Asahi Linux and means the firmware is not imported into the Nix store at evaluation time.
+
+If you do not want the impurity of referencing files on the ESP (or are using flakes where this is prohibited), copy them elsewhere and specify the path explicitly, enabling eval-time extraction:
```
flokli: This is bogus. The described approach just above of mounting / copying from the ESP at runtime means it's not part of eval, so whether you're using flakes or not won't matter.

I guess this should be reworded to something like:

> If you don't want to read from the ESP at boot time, but instead manage firmware with Nix explicitly, use the following snipet. Keep in mind that if also using flakes, the referenced path can't be outside your flake.

docs/uefi-standalone.md
```diff
@@ -243,14 +243,15 @@ Add the `./apple-silicon-support` directory to the imports list and switch off t
...
+Various non-free non-redistributable peripheral firmware files are required to use system hardware like Wi-Fi. The Asahi Linux installer grabs these from macOS and stores them on the EFI system partition when it is created. The NixOS installer loads them from there while booting so that all hardware is available during installation.
 
-Currently, the only supported way to update the peripheral firmware files is to destroy and re-create the EFI system partition, so they will not change unexpectedly. If you do not want the impurity of referencing them (or are using flakes where this is prohibited), copy them off the EFI system partition (e.g. on the installation ISO `mkdir -p /mnt/etc/nixos/firmware && cp /mnt/boot/asahi/{all_firmware.tar.gz,kernelcache*} /mnt/etc/nixos/firmware`) and specify this path in your configuration:
+By default, the Apple Silicon support module now loads peripheral firmware from the ESP at boot time (before any drivers that need it are loaded). This matches the approach used by Fedora Asahi Linux and means the firmware is not imported into the Nix store at evaluation time.
```
flokli:
```suggestion
By default, the Apple Silicon support module now loads peripheral firmware from the ESP at boot time (before any drivers that need it are loaded). This matches the approach used by Fedora Asahi Linux and means the firmware is not imported into the Nix store at evaluation time. It also means there should be no need for any manual firmware management, as in case any more firmware needs to be collected, the Asahi Installer will do it for us, as it does for Fedora.
```

apple-silicon-support/modules/peripheral-firmware/default.nix
```diff
@@ -4,78 +4,242 @@
...
 
-      default = lib.findFirst (path: builtins.pathExists (path + "/all_firmware.tar.gz")) null [
-        # path when the system is operating normally
-        /boot/asahi
-        # path when the system is mounted in the installer
-        /mnt/boot/asahi
-      ];
+      default = lib.findFirst
+        (path:
+          builtins.pathExists (path + "/firmware.cpio")
```
flokli: Why adding support for `firmware.cpio`, in addition to `all_firmware.tar.gz`?

Especially poking around in the filesystem to set a default is a bit too magic.

Let's keep the eval-time firmware support strictly to `all_firmware.tar.gz`.

apple-silicon-support/modules/peripheral-firmware/default.nix
```diff
@@ -4,78 +4,242 @@
...
       type = lib.types.bool;
-      default = true;
+      default = false;
       description = ''
-        Automatically extract the non-free non-redistributable peripheral
-        firmware necessary for features like Wi-Fi.
+        When enabled, automatically extract the non-free non-redistributable
+        Asahi peripheral firmware into the Nix store at evaluation time.
+
+        Enable this if you want declarative/offline firmware management, are
```
flokli: We should copy some of the reasons to why it's actually not preferred to use this in most cases.

And I'd drop the parts about "flakes don't allow you to refer to paths outside your flake". That's normal flake behaviour, nothing special to this option. Having it in uefi-standalone.md is sufficient.

docs/release-notes.md
```diff
@@ -8,6 +8,35 @@ Among others, the kernel has been updated to 6.18.x, and `apple_dcp` was renamed
 to `appledrm`, so users specifying the `apple_dcp.show_notch=1` kernelparam need to
 change it to `appledrm.show_notch=1`.
 
+### Peripheral firmware loading changes
+
+Peripheral firmware (required for Wi-Fi, Bluetooth, etc.) is now loaded from
+the EFI System Partition at boot time by default, rather than being extracted
+into the Nix store at evaluation time. This aligns with the approach used by
+Fedora Asahi Linux and removes eval-time impurities.
+
```
flokli, line 17: This is a bit of a confusing summary.

`peripheralFirmwareDirectory` behaviour should not have changed at all.

docs/release-notes.md
```diff
@@ -8,6 +8,35 @@ Among others, the kernel has been updated to 6.18.x, and `apple_dcp` was renamed
...
+  automatically on the next boot.
+- If you have an older installation with the legacy `asahi/all_firmware.tar.gz`
+  format, you must either:
+  1. Re-run the Asahi Linux installer to update the ESP firmware to the new
+     `vendorfw` format (recommended), or
+  2. Explicitly set `hardware.asahi.extractPeripheralFirmware = true` and
+     `hardware.asahi.peripheralFirmwareDirectory = /boot/asahi` to continue
+     using the legacy eval-time extraction path.
+
+Users who prefer declarative/offline firmware management (e.g. flake users with
```
flokli: Flake users are not required to do anything just because they use flakes, as explained above a few times.

apple-silicon-support/modules/peripheral-firmware/default.nix
Solidsilver: Got it, done.

autrimpo: Does the new firmware update script create an updated `all_firmware.tar.gz`? From my testing it does not, only `firmware.cpio` in `/boot/vendorfw`.

apple-silicon-support/modules/peripheral-firmware/default.nix
```diff
@@ -4,78 +4,244 @@
   lib,
   ...
 }:
-{
-  config = lib.mkIf config.hardware.asahi.enable {
-    assertions = lib.mkIf config.hardware.asahi.extractPeripheralFirmware [
+
+let
+  cfg = config.hardware.asahi;
+  bootTimeFirmware = !cfg.extractPeripheralFirmware;
```
flokli: Drop this and simply mkIf  !cfg.extractPeripheralFirmware at the only usage site.

apple-silicon-support/modules/peripheral-firmware/default.nix
```diff
@@ -4,78 +4,244 @@
...
-              mv vendorfw/* $out/lib/firmware
+      ''
+        mkdir -p $out/lib/firmware
+
+        if [ -f "${firmwareDir}/firmware.cpio" ]; then
+          cpio_src="${firmwareDir}/firmware.cpio"
+
+        elif [ -f "${firmwareDir}/all_firmware.tar.gz" ]; then
+          mkdir extracted
+          ${pkgs'.asahi-fwextract}/bin/asahi-fwextract "${firmwareDir}" extracted
```
flokli: You already add asahi-fwextract to `nativeBuildInputs`. Why change it to an interpolated store path here?

apple-silicon-support/modules/peripheral-firmware/default.nix
```diff
@@ -4,78 +4,244 @@
...
-              mkdir extracted
-              asahi-fwextract ${config.hardware.asahi.peripheralFirmwareDirectory} extracted
-
-              mkdir -p $out/lib/firmware
-              cat extracted/firmware.cpio | cpio -id --quiet --no-absolute-filenames
-              mv vendorfw/* $out/lib/firmware
+      ''
+        mkdir -p $out/lib/firmware
+
+        if [ -f "${firmwareDir}/firmware.cpio" ]; then
```
flokli, line 111: The nixos option states only `all_firmware.tar.gz` is supported, but this seems to work with a `firmware.cpio` in `${firmwareDir}` as well. That's probably ok, though the comment describing this script should probably explain it.

apple-silicon-support/modules/peripheral-firmware/default.nix
```diff
@@ -4,78 +4,244 @@
...
+        ];
+      }
+
+      (lib.mkIf cfg.extractPeripheralFirmware {
+        hardware.firmware = lib.mkIf (cfg.peripheralFirmwareDirectory != null) [
+          asahiPeripheralFirmware
+        ];
+      })
+
+      (lib.mkIf bootTimeFirmware {
```
flokli:
```suggestion
      (lib.mkIf !cfg.extractPeripheralFirmware {
```

Not using that alias makes it much more clear this is an either or :)

iso-configuration/installer-configuration.nix
flokli: The Asahi Installer parts should populate this now, so can drop all this, no?

By the time we boot this, we should already have vendorfw extracted by the Asahi Installer.

flokli: drop entirely?

apple-silicon-support/modules/peripheral-firmware/default.nix
```diff
@@ -4,78 +4,244 @@
...
+
+        The default (disabled) loads firmware from the ESP at boot time, which
+        is the recommended approach and matches upstream Fedora Asahi Linux.
       '';
     };
 
     peripheralFirmwareDirectory = lib.mkOption {
       type = lib.types.nullOr lib.types.path;
 
       default = lib.findFirst (path: builtins.pathExists (path + "/all_firmware.tar.gz")) null [
```
flokli: default can be just null.

apple-silicon-support/modules/peripheral-firmware/default.nix
```diff
@@ -4,78 +4,244 @@
...
+      fi
+    elif [ -f /tmp/asahi-esp/asahi/all_firmware.tar.gz ]; then
+      echo "Warning: legacy all_firmware.tar.gz found on ESP but boot-time extraction only supports vendorfw/firmware.cpio"
+    else
+      echo "Warning: vendorfw/firmware.cpio not found on ESP"
+    fi
+    exit 0
+  '';
+
+  asahiPeripheralFirmware =
```
flokli: Add a comment on top that this is the eval-time firmware extracting code. Maybe it's even cleaner to move it entirely into its own callsite, so it's clearly inside the eval-time firmware extraction if branch.

apple-silicon-support/modules/peripheral-firmware/default.nix
```diff
@@ -4,78 +4,244 @@
...
+        echo "ESP already mounted at $existing, bind-mounting for firmware extraction"
+      else
+        echo "Warning: failed to mount ESP, skipping firmware extraction"
+        rmdir /tmp/asahi-esp 2>/dev/null || true
+        exit 0
+      fi
+    fi
+
+    cleanup() {
+      rm -rf /tmp/asahi-fwextract
```
flokli: `switch_root` should free all this without the need to explicitly cleanup manually.

apple-silicon-support/modules/peripheral-firmware/default.nix
Solidsilver: Good point!

apple-silicon-support/modules/peripheral-firmware/default.nix
Solidsilver: I didn't know this. Did some research into how `switch_root` works. It looks like it does lazy unmounts during cleanup. I can remove `rm -rf /tmp/asahi-fwextract` from this cleanup script - but I think it might be worth keeping the esp unmount to prevent any potential race conditions.

apple-silicon-support/modules/peripheral-firmware/default.nix
Solidsilver: Once the default is `null`,  the `extractPeripheralFirmware` option is no longer needed (logically). The new logic could be:
If `peripheralFirmwareDirectory` = `null` (default), we use boot-time extraction.
if `peripheralFirmwareDirectory` is set to a path, we extract during eval.

Would it be ok if I removed `extractPeripheralFirmware` and added something like
```nix
(lib.mkRemovedOptionModule [ "hardware" "asahi" "extractPeripheralFirmware" ] ''
      Eval-time firmware extraction is now enabled by setting
      `hardware.asahi.peripheralFirmwareDirectory` to a path.
      Leave it as `null` (the default) to load firmware from the ESP at boot time.
    '')
```

flokli: Ok, sgtm! Maybe add a "(recommended)" at the end of the message.

iso-configuration/installer-configuration.nix
Solidsilver: I think this is correct, but I'm not familiar enough with the installer process to be sure. Worth testing.

apple-silicon-support/modules/peripheral-firmware/default.nix
```diff
@@ -4,79 +4,225 @@
...
+  # at eval time (not boot-time)
+  extractAsahiFirmwareAtEval =
+    let
+      pkgs' = cfg.pkgs;
+      firmwareDir = cfg.peripheralFirmwareDirectory;
+    in
+    pkgs.runCommand "asahi-peripheral-firmware"
+      {
+        nativeBuildInputs = [
+          pkgs'.asahi-fwextract
```
flokli:
```suggestion
          cfg.pkgs.asahi-fwextract
```

(and remove the `pkgs'` in the let binding above)

docs/release-notes.md
```diff
@@ -8,6 +8,38 @@ Among others, the kernel has been updated to 6.18.x, and `apple_dcp` was renamed
...
 change it to `appledrm.show_notch=1`.
 
+### Peripheral firmware loading changes
+
+Peripheral firmware (required for Wi-Fi, Bluetooth, etc.) is now loaded from
+the EFI System Partition at boot time by default, rather than being extracted
+into the Nix store at evaluation time. This aligns with the approach used by
+Fedora Asahi Linux.
+
+- `hardware.asahi.extractPeripheralFirmware` has been removed. Eval-time
```
flokli:
```suggestion
- The `hardware.asahi.extractPeripheralFirmware` NixOS option has been removed. Eval-time
```

docs/release-notes.md
```diff
@@ -8,6 +8,38 @@ Among others, the kernel has been updated to 6.18.x, and `apple_dcp` was renamed
...
+  firmware extraction is now controlled solely by
```
flokli:
```suggestion
  firmware extraction is now enabled solely by
```

docs/release-notes.md
```diff
@@ -8,6 +8,38 @@ Among others, the kernel has been updated to 6.18.x, and `apple_dcp` was renamed
...
+  `hardware.asahi.peripheralFirmwareDirectory`.
```
flokli:
```suggestion
  `hardware.asahi.peripheralFirmwareDirectory` set to a path containing `firmware.cpio` or `all_firmware.tar.gz`.
```

docs/release-notes.md
```diff
@@ -8,6 +8,38 @@ Among others, the kernel has been updated to 6.18.x, and `apple_dcp` was renamed
...
+- `hardware.asahi.peripheralFirmwareDirectory` now defaults to `null`. When
```
flokli: It defaults to `null`, so boot time extraction of firmware is the new recommended default.

docs/release-notes.md
```diff
@@ -8,6 +8,38 @@ Among others, the kernel has been updated to 6.18.x, and `apple_dcp` was renamed
...
+  set to a path, firmware is extracted at evaluation time. When left as
+  `null` (the default), firmware is loaded from the ESP at boot time.
```
flokli:
```suggestion
```

docs/release-notes.md
```diff
@@ -8,6 +8,38 @@ Among others, the kernel has been updated to 6.18.x, and `apple_dcp` was renamed
...
+     `vendorfw` format (recommended), or
+  2. Set `hardware.asahi.peripheralFirmwareDirectory` to the directory
+     containing your legacy firmware files (e.g. `/boot/asahi`) to continue
+     using eval-time extraction.
+
+Users who prefer declarative/offline firmware management can opt into
+eval-time extraction by setting `hardware.asahi.peripheralFirmwareDirectory`
+to a local copy of the firmware. Both the modern `firmware.cpio` format
+(copy it from `/boot/vendorfw`) and the legacy `all_firmware.tar.gz` format
+are supported.
```
flokli:
```suggestion
and the legacy `all_firmware.tar.gz` format are supported.
```

Copying is only necessary if flakes are used, and not being able to reference to dirs outside your flake is something flake users are already accustomed to.

This is nothing new - they either already did copy to firmware into their flake, or as a new installation they can use boot time extraction or follow the uefi-standalone.md docs which contain the hint for flakes users.

docs/uefi-standalone.md
```diff
@@ -243,16 +243,19 @@ Add the `./apple-silicon-support` directory to the imports list and switch off t
...
+
+If you prefer to manage firmware declaratively in your Nix configuration rather than reading from the ESP at boot time, set the firmware path:
 ```
-  # Specify path to peripheral firmware files.
   hardware.asahi.peripheralFirmwareDirectory = ./firmware;
-  # Or disable extraction and management of them completely.
-  # hardware.asahi.extractPeripheralFirmware = false;
 ```
 
+Both `firmware.cpio` (from `/boot/vendorfw` on a modern installation) and the legacy `all_firmware.tar.gz` format are supported for eval-time extraction. Place the firmware file(s) in the referenced directory.
```
flokli:
```suggestion
Both `firmware.cpio` (from `/boot/vendorfw` on a modern installation) and the legacy `all_firmware.tar.gz` format are supported for eval-time extraction.
Point the path to a location where the either of the two files reside.
```

non-flakes users wanting to use eval-time firmware extraction can absoultely point to `/boot/vendorfw`.

docs/uefi-standalone.md
```diff
@@ -243,16 +243,19 @@ Add the `./apple-silicon-support` directory to the imports list and switch off t
...
+
+Keep in mind that if also using flakes, the referenced path can't be outside your flake.
```
flokli:
```suggestion
Keep in mind that if also using flakes, the referenced path can't be outside your flake, so you might need to copy them in.
```

flokli (commented):
Please squash the history a bit.

docs/release-notes.md
```diff
@@ -8,6 +8,38 @@ Among others, the kernel has been updated to 6.18.x, and `apple_dcp` was renamed
...
+
+- `hardware.asahi.extractPeripheralFirmware` has been removed. Eval-time
+  firmware extraction is now controlled solely by
+  `hardware.asahi.peripheralFirmwareDirectory`.
+- `hardware.asahi.peripheralFirmwareDirectory` now defaults to `null`. When
+  set to a path, firmware is extracted at evaluation time. When left as
+  `null` (the default), firmware is loaded from the ESP at boot time.
+
+**Action required for existing installations:**
+- If your ESP already has the `vendorfw/firmware.cpio` format (Asahi installer
```
flokli: There's no change required if the user is currently using eval-time firmware extraction, but as the boot-time extraction is the default for new installations, users should consider to switch over.

We should also document how to recreate / update the vendor firmware package, if they don't have the right format yet, or potentially miss some firmware:

> Invoke the Asahi Installer by running curl https://alx.sh/ | sh
> Choose the “Rebuild vendor firmware package” option when prompted

apple-silicon-support/modules/peripheral-firmware/default.nix
```diff
@@ -4,79 +4,225 @@
   lib,
   ...
 }:
+
+let
+  cfg = config.hardware.asahi;
+
+  # Script to extract Asahi firmware
+  # during boot
```
flokli:
```suggestion
  # Script to extract Asahi firmware during boot
```

apple-silicon-support/modules/peripheral-firmware/default.nix
```diff
@@ -4,79 +4,225 @@
...
+  extractAsahiFirmwareInitrdScript = pkgs.writeShellScript "asahi-firmware-extract-initrd" ''
+    set -u
+    mkdir -p /run/asahi-firmware
+
+    # m1n1 may stage vendor firmware directly into the initramfs as /vendorfw.
```
flokli: Where is this coming from? Can you point to the code/documentation describing where this happens?

flokli: So this is not m1n1, but the bootloader (in the case of NixOS, usually sd-boot) loading the vendorfw as another initrd. We don't expose this functionality in NixOS yet, so it can probably be removed for now.

Solidsilver: Oh, good to know. Will do.

docs/uefi-standalone.md
```diff
@@ -247,16 +247,14 @@ Various non-free non-redistributable peripheral firmware files are required to u
...
-If you prefer to manage firmware declaratively in your Nix configuration rather than reading from the ESP at boot time, enable eval-time extraction and specify the path manually:
+If you prefer to manage firmware declaratively in your Nix configuration rather than reading from the ESP at boot time, set the firmware path:
 ```
-  # Enable eval-time extraction and specify the path.
-  hardware.asahi.extractPeripheralFirmware = true;
   hardware.asahi.peripheralFirmwareDirectory = ./firmware;
 ```
 
-Both `firmware.cpio` (from `/boot/vendorfw` on a modern installation) and the legacy `all_firmware.tar.gz` format are supported for eval-time extraction. Place the firmware file(s) in the referenced directory.
+Both `firmware.cpio` (from `/boot/vendorfw` on a modern installation) and the legacy `all_firmware.tar.gz` format are supported for eval-time extraction. Point the path to a location where the either of the two files reside.
```
flokli: s/the either/either/

flokli:
I'm sorry this is taking so many iterations, but there might still be more work left here.

This currently causes *only* the asahi firmware to be loadable, but whatever the user / NixOS sets `hardware.firmware` to is not accessible anymore. It would not be a problem if none of the firmware would be needed, but there's firmware in there used for various peripherals that the user can connect via USB as well.

To migitate that, we would need to setup an overlayfs mount, consisting of the unpacked vendor firmware and what config.firmware is pointing to.

Would you be up to extending this PR?

Solidsilver:
> This currently causes only the asahi firmware to be loadable, but whatever the user / NixOS sets hardware.firmware to is not accessible anymore.

Ah, that totally makes sense. As I was implementing that part I had a little feeling in the back of my head that something could be missed with how it forces loading the asahi stuff at boot.

Yep, no worries! I want it to be done right. I'll take a look into  in the next couple days.

Solidsilver:
@flokli I've made the requested changes. I don't have any other firmware defined, so I haven't actually tested that both firmware load. But the overlay looks to be created correctly on boot.

yuyuyureka (commented):
Please leave an option to disable all kinds of automatic firmware extraction

flokli:
> Please leave an option to disable all kinds of automatic firmware extraction

When setting `hardware.asahi.peripheralFirmwareDirectory` to a path containing the firmware, it'll happen at build time. You mean also keeping the boolean around, and then trying to boot without any firmware at all?

yuyuyureka:
> You mean also keeping the boolean around, and then trying to boot without any firmware at all?

Yes, this. These scripts look way too complicated for what I need, and I want to just disable them.

Solidsilver:
> Yes, this. These scripts look way too complicated for what I need, and I want to just disable them.

You're right - the boot-time scripts are indeed more involved than the old eval-time approach.

Just to clarify the behavior in this PR: setting `hardware.asahi.peripheralFirmwareDirectory` to a directory containing `firmware.cpio` or `all_firmware.tar.gz` **disables** the new boot-time ESP/initrd scripts. When using this option, the behavior is effectively the same as before this PR - just with an explicit path instead of auto-detecting `/boot/asahi`.

So I see two separate cases:
1. “I do not want the initrd scripts; I want explicit/declarative firmware management.”
  -> That should already be covered by `hardware.asahi.peripheralFirmwareDirectory`.
2. “I want to use the Apple Silicon support module, but load no Asahi peripheral firmware at all.”
  -> This is what the `extractPeripheralFirmware` option allowed, which would be removed in this PR

Is case 2 the use case you want preserved?

yuyuyureka:
> Is case 2 the use case you want preserved?

Exactly

Solidsilver:
> > Is case 2 the use case you want preserved?
> 
> 
> 
> Exactly

Ok, great. I will put that back in.

Solidsilver:
@flokli @yuyuyureka Is there anything else I can do to help get this merged in?

flokli:
It's mostly me needing to find the time to properly review and test it.

I've also had some conversations in the wider NixOS community, as I still think getting the firmware added to the initrd by just specifying an additional initrd path in the bootloader config would be preferable in the long term, but threading this through all the layers is a bit of work (and probably shouldn't block this PR)

sempiternal-aurora:
if it helps at all, I've been using the new `vendorfw` loading at boot time for the past few months, and have had no issues, everything seems to be working well. Though I also are not using any other firmware other than from `/boot/vendorfw`.

flokli:
So, the manual mounting in the initrd was bugging me more than I want to admit, and I took a stab at this with @r-vdp :laughing: 

There's now https://github.com/NixOS/nixpkgs/pull/534265, which adds a `system.boot.additionalInitrd.paths` option to NixOS, and adds support for this via a bootspec extension. This is supported for sd-boot and limine (once the already-landed fix to Limine trickled into a release and into nixpkgs).

flokli:
So with #494 and #496 we're using the Asahi Installer-provided format.

We're still not loading it at runtime via the bootloader, but with the current `main` branch ALS should work out of the box, provided you follow the instructions in the release notes (or installer docs for new installations).

Mounting the vendorfw from the ESP at bootup is still something I'd like to do in the future, but probably after a release, which has been long overdue.

As far as this PR is concerned, I spotted some more things I'd like to change, and explaining what to do would be more churn than just proposing the change by myself.

I suggest we close this PR, do a release and I'll open another PR incorporating some of this logic. Still thanks for your contribution! It helped to shape all this.

PR #453, opened 2026-04-30
linux-asahi: 6.19.12-1 -> 6.19.14-1

yuyuyureka:


flokli:
Successfully switched my m2air to this.

Apart from the copy.fail fixes it includes https://github.com/AsahiLinux/linux/pull/490, so now the `IIO_AOP_SENSOR_ALS` module is compiled.

It still yells at me because I didn't provide the firmware / calibration for it, but it's a step forward and https://github.com/nix-community/nixos-apple-silicon/pull/451 should deal with that :)

PR #464, opened 2026-05-15
linux-asahi: 6.19.14-2 -> 7.0.8-1

yuyuyureka:


reinthal:
https://asahilinux.org/2026/04/progress-report-7-0/

for those looking for the update notes

elisaado:
checked out this branch, compiled & used the 7.0 kernel both with and without fairydust patches, seems to be stable.

PR #468, opened 2026-05-18
linux-asahi: 7.0.8-1 -> 7.0.9-2

yuyuyureka:


elisaado:
Might want to bump to [7.0.10-2](https://github.com/AsahiLinux/linux/releases/tag/asahi-7.0.10-2) in one release instead?

elisaado:
see #471

PR #470, opened 2026-05-20
linux-asahi: allow user to pass extra arguments to buildLinux

imnotpoz:
without this I couldn't find a way to override things like `enableCommonConfig` or `ignoreConfigErrors` in the kernel package - I'd be happy to use that if there is a way, although this still seems like a more convenient option


flokli:
@imnotpoz this needs a rebase.

imnotpoz:
fixed the merge conflict

flokli:
Sorry, I forgot about this again. Just rebased, sorry for the delay.

PR #471, opened 2026-05-24
linux-asahi: 7.0.8-1 -> 7.0.10-2

elisaado:
Tested on 16' M1 MBP with fairydust patches, seems to boot & run like it should


elisaado:
(amended the commit message to match the ones in this repo)

PR #475, opened 2026-06-03
linux-asahi: asahi-7.0.10-2 -> asahi-7.0.11-1

elisaado:
not tested yet


elisaado:
Tested with fairydust patches, everything still works

PR #429, opened 2026-03-02
Add limine support

anshulnoori:
Limine support 👍

From [search.nixos.org](https://search.nixos.org/options?channel=unstable&query=boot.loader.limine.&show=boot.loader.limine.additionalFiles):
> [boot.loader.limine.additionalFiles](https://search.nixos.org/options?channel=unstable&query=boot.loader.limine.&show=boot.loader.limine.additionalFiles)
> Name: boot.loader.limine.additionalFiles
> Description:
> A set of files to be copied to /boot. Each attribute name denotes the destination file name in /boot, while the corresponding attribute value specifies the source file.
> Type: attribute set of absolute path
> Default: { }
> Example: `{ "efi/memtest86/memtest86.efi" = "${pkgs.memtest86-efi}/BOOTX64.efi"; }`
> Declared in: [nixos/modules/system/boot/loader/limine/limine.nix](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/nixos/modules/system/boot/loader/limine/limine.nix)

PR #480, opened 2026-06-08
modules/boot-m1n1: install m1n1 stage 2 via systemd-tmpfiles

### LINKED ISSUE #440: Usb seems to be broken in macbook M2
So, it has been some time that I noticed that the usb ports of my macbook M2 was not working in NixOS, any usb device that I added would not be recognized at all.

Yesterday, I found this command in a reddit post:

```sh
nix build .#nixosConfigurations.YOURHOSTNAME.config.system.build.m1n1
```

This will generate a new `boot.bin` file. I replaced my original one (`/boot/m1n1/boot.bin`) with this one and restarted the machine and now USB is working again just fine.

Is this a bug? I noticed that my old `boot.bin` field was modified in 01 Jan, why isn't nixos updating that file automatically? Do I need to add some code to my NixOS configuration to make it do that automatically?

eConnah:
*(Edited)*

Description:
Limine sandboxes extraFiles, so the m1n1 binary wasn't being placed where m1n1 stage 1 expects it, and the file wasn't updating on rebuilds. This replaces the bootloader-specific extraFiles logic with systemd.tmpfiles.rules that copy the binary directly to /boot/m1n1/boot.bin, matching the m1n1 User Guide. The copy happens immediately during nixos-rebuild switch, and also at boot.

Changes:
- Removed boot.loader.*.extraFiles lines.
- Added systemd.tmpfiles.rules to manage /m1n1/boot.bin on the ESP.

Testing:
- Deleted the existing m1n1 directory and rebuilt -> file reappeared instantly.
- Rebooted successfully with the new binary.

Fixes #440

### DIFF
--- a/apple-silicon-support/modules/boot-m1n1/default.nix
+++ b/apple-silicon-support/modules/boot-m1n1/default.nix
@@ -33,10 +33,12 @@ let
 in
 {
   config = lib.mkIf config.hardware.asahi.enable {
-    # install m1n1 with the boot loader
-    boot.loader.grub.extraFiles = bootFiles;
-    boot.loader.systemd-boot.extraFiles = bootFiles;
-    boot.loader.limine.additionalFiles = bootFiles;
+    systemd.tmpfiles.rules = [
+      # Remove the old version first so it gets replaced every time
+      "r ${config.boot.loader.efi.efiSysMountPoint}/m1n1/boot.bin"
+      # Copy the new one
+      "C ${config.boot.loader.efi.efiSysMountPoint}/m1n1/boot.bin 0644 root root - ${bootFiles."m1n1/boot.bin"}"
+    ];
 
     # ensure the installer has m1n1 in the image
     system.extraDependencies = lib.mkForce [


apple-silicon-support/modules/boot-m1n1/default.nix
```diff
@@ -33,10 +33,14 @@ let
 in
 {
   config = lib.mkIf config.hardware.asahi.enable {
-    # install m1n1 with the boot loader
-    boot.loader.grub.extraFiles = bootFiles;
-    boot.loader.systemd-boot.extraFiles = bootFiles;
-    boot.loader.limine.additionalFiles = bootFiles;
+    system.activationScripts.installM1n1 = {
```
flokli: We shouldn't be adding more activation scripts. See https://github.com/NixOS/nixpkgs/issues/475305 for context.

Is this only an issue with Limine? https://github.com/nix-community/nixos-apple-silicon/pull/429#issuecomment-3999442001 states that Limine isn't really well-supported/tested.

Skimming through `nixos/modules/system/boot/loader/limine/limine.nix`, I don't really see where the sandboxing is happening that you're mentioning.

flokli (changes requested):

apple-silicon-support/modules/boot-m1n1/default.nix
eConnah: It uses its own folder here which is where the extraFiles places m1n1 theres no way to go up a folder. Is there a way this can be done without an activation script?

image
image

eConnah: the one in the correct location is the old one and doesn't update anymore

flokli: You could use a oneshot systemd service that copies it to the right location, if limine is enabled (alongside a comment why we cannot use `boot.loader.limine.additionalFiles` instead.

eConnah: understood should i keep grub and systemd the same as before or just make everything use the oneshot service? I assume the one shot service will work for any boot manager.

eConnah:
I deleted my m1n1, then updated my flake input and rebuilt, the file got recreated and I succeeded in rebooting. I did it this way so it should work when m1n1 updates before the reboot not on next boot.

```
# Before reboot
  File: /boot/m1n1/boot.bin
  Size: 7737827   	Blocks: 15120      IO Block: 4096   regular file
Device: 259,4	Inode: 8           Links: 1
Access: (0600/-rw-------)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2026-06-08 02:00:00.000000000 +0200
Modify: 1980-01-01 01:00:00.000000000 +0100
Change: 1980-01-01 01:00:00.000000000 +0100
 Birth: 2026-06-10 14:00:11.570000000 +0200

# After reboot before rebuild switch
  File: /boot/m1n1/boot.bin
  Size: 7737827   	Blocks: 15120      IO Block: 4096   regular file
Device: 259,4	Inode: 8           Links: 1
Access: (0600/-rw-------)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2026-06-08 02:00:00.000000000 +0200
Modify: 1980-01-01 01:00:00.000000000 +0100
Change: 1980-01-01 01:00:00.000000000 +0100
 Birth: 2026-06-10 14:02:09.720000000 +0200

# After rebuild switch
   File: /boot/m1n1/boot.bin
  Size: 7737827   	Blocks: 15120      IO Block: 4096   regular file
Device: 259,4	Inode: 9           Links: 1
Access: (0600/-rw-------)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2026-06-08 02:00:00.000000000 +0200
Modify: 1980-01-01 01:00:00.000000000 +0100
Change: 1980-01-01 01:00:00.000000000 +0100
 Birth: 2026-06-10 14:04:04.980000000 +0200

```

flokli:
I rebased the commit, updated the commit message and dropped the "d " directive, as:

> Note  that for all line types that result in creation of any kind of file node (i.e.  f,
> d/D/v/q/Q, p, L, c/b and C) leading directories are implicitly created if needed,  owned
> by  root  with  an  access mode of 0755. In order to create them with different modes or
> ownership make sure to add appropriate d lines.

Let me know if this works for you, happy to merge soon.

flokli:
I opened https://github.com/NixOS/nixpkgs/issues/537466 as a nixpkgs bug. Ultimately, in the long term, we might still want to use these options, or at least prevent other people from running into the same issues...

flokli (approved):

eConnah:
Just tested it and it works perfectly for me tysm.

flokli:
Seems this does /not/ work reliably, as reported. Someone please fix https://github.com/NixOS/nixpkgs/issues/537466 in nixpkgs, then this should work for Limine too.

eConnah:
~~I was reading the irc and someone said they didnt have an m1n1 folder at all which is strange as my tmpfiles only removes the file not the folder~~

~~https://oftc.catirclogs.org/asahi/2026-07-03#35469264~~

~~but could the issue be due to the removal of the directory step and it worked for me as the directory had been created during testing?~~
nvm I tested it again with a folder called m2n2 and made sure that it didnt exist in my /boot before I rebuilt but it still placed the boot.bin there so I am unaware of whats causing this.

also for anyone that needs the instructions for a fix here, again my apologies :(
make another nixos bootable usb, boot into it, mount your boot partition and data partition.
then
```
sudo find /mnt/nix/store/ -name "*boot.bin"
/nix/store/0cpjbwwp493b35a1mqz38kaxmjrb4r1f-uboot-apple_m1_defconfig-2026.04-2-asahi/m1n1-u-boot.bin
/nix/store/a5rri199zziikd4vnsk2i2zip1lq1y25-boot.bin
```
its the normal boot.bin one here if you have multiple check the dates using stat, then copy the file to /mnt/boot/m1n1/boot.bin.
Then just unmount and shutdown unplug the bootable usb and then boot as normal, then you must update your flake apple-silicon support input and rebuild. If you also feel like it any logs would be super helpful to figure out what went wrong for people.

flokli:
There's a proper fix for this nixpkgs bug sitting in https://github.com/NixOS/nixpkgs/pull/539368.

Even though there's already a regression test, it would be great if some Asahi Limine users could also take a look, and test-drive this PR with current n-a-s main, and comment there.

eConnah:
Seems it got merged before I could test it however I just did it right now and it worked perfectly, tysm man.

PR #492, opened 2026-07-06
gh actions: drop namespacelabs/nscloud-cache-action

flokli:
Looking at https://github.com/namespace-integration-demos/nix-cache-example/blob/main/.github/workflows/demo.yaml, you're supposed to put `path: /nix` in there, not ~/.cache/nix.

Having a cache that maybe contains (stale) data is scary, we really don't want that, at least not if we also don't also keep the rest of that state.

### DIFF
--- a/.github/workflows/cache.yml
+++ b/.github/workflows/cache.yml
@@ -35,9 +35,6 @@ jobs:
       uses: actions/checkout@v7
       with:
         ref: "${{ matrix.ref }}"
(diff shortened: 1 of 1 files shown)


flokli:
checked back with @yuyuyureka, this seems to be mostly client-side fetcher things, so not depending on any state of /nix/store or the nix db. At leas not in the current Nix version ;-)

Issue #500, opened 2026-07-08
Allow disabling `hardware.asahi.enable`


flokli:
This currently is not possible, as some code unconditionally accesses option values even if when `hardware.asahi.enable` is set to false, showing errors like this:

> error: The option `hardware.asahi.pkgs' was accessed but has no value defined. Try setting the option.

We need to chase down these occurences, and gate them on `hardware.asahi.enable` being set to true.

PR #501, opened 2026-07-08
cleanup hardware.asahi.pkgs usages

flokli:
`asahi-fwextract` and `m1n1` live in nixpkgs now, no need to pull in our custom pkgs instance.

Make sure the remaining usages are inside `cfg.enabled` blocks, fixing #500.

### DIFF
--- a/apple-silicon-support/modules/boot-m1n1/default.nix
+++ b/apple-silicon-support/modules/boot-m1n1/default.nix
@@ -4,43 +4,46 @@
   lib,
   ...
 }:
-let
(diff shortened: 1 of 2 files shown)


flokli:
I was successfully able to evaluate my system with this set to false.

Didn't switch to it on my m2air obviously, as it would remove m1n1 etc, but this now evalutated if you set it to false as well.

Meaning people could start including the module on all their systems unconditionally, once we flip the default (which would happen in the release after the upcoming one).

yuyuyureka (approved):

PR #502, opened 2026-07-10
docs/uefi-standalone: fix doubled word in proxyclient section

KangaZero:
Remove the duplicated "on" in the sentence describing proxyclient usage on x86_64.

Before: "To use some of the proxyclient features on on x86\_64"
After:  "To use some of the proxyclient features on x86\_64"


flokli:
Thanks!

Issue #417, opened 2026-01-17
Notch suddenly doesn't work anymore


wtestcase:
Hello, I have my kernel boot parameters set as follows to "enable" the notch:
`boot.kernelParams = ["apple_dcp.show_notch=1"];`
This has always worked fine, but now, it suddenly doesn't anymore, the notch's part of the screen is simply black. I'd be glad if somebody could help.

flokli:
Please check https://github.com/nix-community/nixos-apple-silicon/commit/940dd182b639f3f26d0b20430f71b63f3b818bdd, this is documented in the release notes.

wtestcase:
Thank you, missed that, sorry

PR #512, opened 2026-07-19
modules/default.nix: create notch enable option

utopiatopia:
When the notch option changed, I saw people on the subreddit, as well as here (#417) saying that the notch doesn't work. Perhaps creating an option that enables the notch for users is warranted?

LMK if this doesn't fit this repo, the description needs changing, or the option needs to be moved to a different submodule (i.e. kernel).

### DIFF
--- a/apple-silicon-support/modules/default.nix
+++ b/apple-silicon-support/modules/default.nix
@@ -30,6 +30,8 @@ in
       else
         pkgs;
 
+    boot.kernelParams = lib.mkIf cfg.notch.enable [ "appledrm.show_notch=1" ];
+
     # 900 is higher priority than mkDefault but lower than just setting
     hardware.sensor.iio.enable = lib.mkOverride 900 true;
   };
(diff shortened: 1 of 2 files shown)


flokli:
There's some movement coming from the phosh ecosystem to actually describe cutouts / notches, and expose this to compositors via a wayland protocol: https://phosh.mobi/posts/notch-support/. 

This recently ended up in phoc: https://gitlab.gnome.org/World/Phosh/phoc/-/merge_requests/734.

I discussed this on [IRC](https://oftc.catirclogs.org/asahi-dev/2026-07-31#1785496080-1785496171;) today, it might be we switch to exposing the whole screen by default, once this sufficiently propagated into downstream compositors. And at which point there would be an option to opt out, rather than opt in. 

So I don't feel comfortable adding a NixOS option for it, and committing to it. The default is sensible, and it's reasonable for people who want to use the area next to the notch today to follow these developments (summarized in the nixos-apple-silicon release notes).

PR #544, opened 2026-09-09
modules/video: add avd support module

sempiternal-aurora:
Add an option under `hardware.asahi.avd` to setup the Apple Video Decoder for use, as well as an option to use `libva-v4l2_request` to provide a vaapi wrapper for the hardware module. I have tested, and the nixpkgs bump doesn't seem to have the python3Packages.dbus-python cross compilation problems. I built it cross compiling on the latest nixpkgs, and I am currently manually built the `installer-bootstrap` on my framework, no errors so far.

It also has built and works on my t8112-j413, however when using `mpv`, the command needs to be `mpv --hwdec=vaapi --vo=dmabuf-wayland <video file>` because of some problem with gpu-next? According to chadmed. Because of these problems, it is marked as experimental and disabled by default.

### DIFF
--- a/apple-silicon-support/modules/video/default.nix
+++ b/apple-silicon-support/modules/video/default.nix
@@ -13,6 +13,17 @@ in
     hardware.firmware = [
       pkgs.avd-fw
     ];
+
+    hardware.graphics = lib.mkIf (cfg.avd.vaapi-support) {
+      enable = true;
+      extraPackages = [
+        pkgs.libva-v4l2_request-sofus13
+      ];
+    };
+
+    environment.sessionVariables = lib.mkIf (cfg.avd.vaapi-support) {
+      LIBVA_DRIVER_NAME = "v4l2_request-sofus13";
+    };
   };
 
   options.hardware.asahi.avd = {
@@ -21,6 +32,13 @@ in
       default = config.hardware.asahi.enable;
       description = ''
         Setup the firmware required for the Apple Video Decoder to work properly.
+      '';
+    };
+    vaapi-support = lib.mkOption {
+      type = lib.types.bool;
+      default = false;
+      description = ''
+        Use libva-v4l2_request-sofus13 to expose a vaapi interface with AVD (Experimental)
       '';
     };
   };
--- a/apple-silicon-support/modules/video/default.nix
+++ b/apple-silicon-support/modules/video/default.nix
@@ -1,0 +1,27 @@
+{
+  config,
+  pkgs,
+  lib,
+  ...
+}:
+
+let
+  cfg = config.hardware.asahi;
+in
+{
+  config = lib.mkIf (cfg.enable && cfg.avd.enable) {
+    hardware.firmware = [
+      pkgs.avd-fw
+    ];
+  };
+
+  options.hardware.asahi.avd = {
+    enable = lib.mkOption {
+      type = lib.types.bool;
+      default = config.hardware.asahi.enable;
+      description = ''
+        Setup the firmware required for the Apple Video Decoder to work properly.
+      '';
+    };
+  };
+}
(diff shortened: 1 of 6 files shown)


apple-silicon-support/packages/libva-v4l2_request/default.nix
```diff
@@ -0,0 +1,47 @@
...
+    libdrm
+    linuxHeaders
+  ];
+
+  mesonFlags = [
+    "-Ddriverdir=${placeholder "out"}/lib/dri"
+  ];
+
+  meta = {
+    description = "VA-API (libva) backend driver for V4L2 stateless video decoders using
```
flokli: Please add a longDescription, or comment that this is an asahi-specific fork of `https://xff.cz/git/libva-v4l2_request`, not the bootlin upstream one.

apple-silicon-support/packages/libva-v4l2_request/default.nix
```diff
@@ -0,0 +1,47 @@
...
+  fetchFromGitHub,
+  meson,
+  ninja,
+  pkg-config,
+  libva,
+  libdrm,
+  linuxHeaders,
+}:
+stdenv.mkDerivation {
+  pname = "libva-v4l2_request";
```
flokli:
```suggestion
  pname = "libva-v4l2_request-sofus13";
```

...and also attribute names / file names. We don't want this to conflict with any of the other (possible) upstreams if one of these ends up in nixpkgs.

apple-silicon-support/packages/libva-v4l2_request-sofus13/default.nix
```diff
@@ -0,0 +1,51 @@
+{
```
flokli, line 1: please also add a paragraph to `docs/release-notes.md` (with a dummy `yyyy-mm-dd (unreleased)` header) explaining the introduction of the `hardware.asahi.avd.{enable,vaapi-support}` option(s).

apple-silicon-support/modules/video/default.nix
```diff
@@ -0,0 +1,45 @@
...
+
+  options.hardware.asahi.avd = {
+    enable = lib.mkOption {
+      type = lib.types.bool;
+      default = config.hardware.asahi.enable;
+      description = ''
+        Setup the firmware required for the Apple Video Decoder to work properly.
+      '';
+    };
+    vaapi-support = lib.mkOption {
```
flokli, line 37: I wonder if there's much value in loading the firmware without setting up vaapi.

If there were also vulkan video support, it would maybe make sense to put this behind a separate option, or have some option asking which methods to enable (with a sane default).

But as we only have vaapi via v4l2 right now I feel like we can just unconditionally enable vaapi if the enable option is set (and drop this option).

sempiternal-aurora: I just wasn't sure if we wanted the vaapi support via default, because it is a significant change, and if you don't pass the right options to mpv it can break things, see below:
image
So I kind of just wanted to leave it disabled by default for now.

apple-silicon-support/packages/libva-v4l2_request-sofus13/default.nix
sempiternal-aurora: Ah, documentation, noooo. I've given it a shot, let me know your thoughts.

apple-silicon-support/modules/video/default.nix
flokli: Agreed on not enabling it by default. The idea was more to only have `hardware.asahi.avd.enable` (defaulting to `false`, not to the value of `config.hardware.asahi.enable`), and no additional `hardware.asahi.avd.vaapi-support` option, but `.enable` configuring vaapi /and/ providing AVD firmware.

Because right now the only way to make use of the firmware is by using vaapi, as there's no vulkan video or anything else yet (so the `.enable` option is a bit useless).

flokli: Ah thanks, the release notes actually explain now why this should be a separate flag. Thank you.

PR #548, opened 2026-09-12
linux-asahi: asahi-7.1.13-1 -> asahi-7.1.13-3

flokli:


apple-silicon-support/packages/linux-asahi/default.nix
sempiternal-aurora, line 61:
```suggestion
              APPLE_PMGR_PWRSTATE = yes;
              
              # no by default, needs to be 'y' to ensure bt audio doesn't stutter.
              BT_BRCMEXT = yes
```
Everything else looks good, this is probably just good to set to ensure that BT audio doesn't stutter.

sempiternal-aurora (commented):
Builds and runs on my m2 air and m3 max mbp, can use wpa_supplicant to WPA3 personal. Haven't tested enterprise connections though, but everything seems to work as intended otherwise.

apple-silicon-support/packages/linux-asahi/default.nix
flokli: Thanks! Pushed a commit adding it.

Issue #538, opened 2026-08-30
Read firmware / calibration data from ESP at boot time


flokli:
The Asahi Linux installer drops a `firmware.cpio` file into the `ESP/vendorfw` directory. It contains some non-redistributable firmware, as well as device-specific calibration files.

We currently have `asahi.hardware.peripheralFirmwareDirectory` default to `/boot/vendorfw`, effectively reading the directory containing `firmware.cpio` at eval time, and appending it to `hardware.firmware`.

This causes annoyances for flake users, as you cannot refer to things outside your flake, [requiring to copy and check them into the repository](https://github.com/nix-community/nixos-apple-silicon/pull/520/changes) (which is an issue, as they're non-redistributable). Workarounds as using `--impure`, or `fetchTree` as suggested in https://github.com/nix-community/nixos-apple-silicon/issues/536 are one way to deal with this.

However, as described in https://asahilinux.org/docs/platform/open-os-interop/#linux-specific, we want to switch to reading this .cpio file from the ESP at *boot time only*, either by making it appear in the initramfs root fs by just adding it as an additional initrd (https://github.com/NixOS/nixpkgs/pull/534265, https://github.com/Limine-Bootloader/Limine/pull/599). As a fallback, and for bootloaders / config plumbing not supporting this (looking at you, GRUB), it also needs to fallback to 'manually reading it from the ESP'.

For the "mouting the ESP and retrieving firmware to load" part, `iso-configuration/installer-configuration.nix` already has a slightly cursed version (using `asahi-fwextract instead of the `firmware.cpio` file in `ESP/vendorfw/firmware.cpio`).

It should be updated to use `vendorfw/firmware.cpio`, and made a bit more resilient towards timing issues (and only mount if vendorfw was not populated via an additional initrd). Then we can start using it as a snippet to include in both the installer medium as well as regular NixOS installations.

All of this should happen by default. It should relieve users from having to point `asahi.hardware.peripheralFirmwareDirectory`  to anything and deal with the pain when also using flakes. `asahi.hardware.peripheralFirmwareDirectory`  could then default to null, the new functionality to load the vendorfw at boottime (new option) could be enabled by default.

Users still wanting to use `asahi.hardware.peripheralFirmwareDirectory` or `hardware.firmware` could set it false again. We probably should have an assertion so users shoud explicitly disable the new logic if `asahi.hardware.peripheralFirmwareDirectory` is (still) set.

[...]

PR #550, opened 2026-09-14
Add support for extracting firmware at boot

CodeWitchBella:
Fixes most of #538 (thanks for the write-up)

I took a stab at implementing the "extract firmware at boot ticket". I marked it as experimental because I only tried it on my M1 Mac mini. I don't really use any of the peripheral firmware on that machine, but it seems like the wifi works.

I decided to stick it in stage 2 because I know more about how that works than about stage 1. The main idea is to basically copy the asahi-scripts script and hope it works. Turns out that it does with minor tweaks :tada: 

I didn't touch the installer, nor did I make it the default yet, but it should move us closer to the desired end goal.

References:
 - https://asahilinux.org/docs/platform/open-os-interop/#firmware-provisioning
 - https://github.com/AsahiLinux/asahi-scripts/blob/main/dracut/modules.d/99asahi-firmware/load-asahi-firmware.sh


imnotpoz:
not sure how else I can help but after rebuilding with `hardware.asahi.peripheralFirmwareDirectory = null;` and rebooting everything works on a macbook m2 pro

please let me know if you'd like any more testing, I'd be happy to

flokli:
Hey, thanks. I didn't take too close of a look yet, but can you elaborate why you chose not to run this in the initrd? Any specific obstacles, or simply a 'this was easier to debug'? Would you be up to moving it to the initrd?

I'm asking because https://asahilinux.org/docs/platform/open-os-interop/#linux-specific specifically describes this as happening in the initrd, and only there we can also make use of the cpio files provided by the bootloader and skip having to mount the ESP to get it from there in (most) cases.

apple-silicon-support/modules/peripheral-firmware/default.nix
```diff
@@ -4,17 +4,107 @@
...
+          "shutdown.target"
+        ];
+        conflicts = [ "shutdown.target" ];
+        wantedBy = [ "sysinit.target" ];
+
+        unitConfig = {
+          DefaultDependencies = false;
+          ConditionPathExists = [
+            "/proc/device-tree/chosen/asahi,efi-system-partition"
+            # already loaded, either by the bootloader or by an earlier start
```
flokli, line 48:
```suggestion
            # already loaded, either by the bootloader or by an earlier run of this service
```

"start" might be misinterpretable as "boot", but "/lib/firmware/vendor" is a tmpfs.

apple-silicon-support/modules/peripheral-firmware/default.nix
```diff
@@ -4,17 +4,107 @@
...
+
+            # mount_sys_esp from asahi-scripts' functions.sh
+            mountpoint=/run/.system-efi
+            mkdir -p "$mountpoint"
+            while grep -q " $mountpoint " /proc/mounts; do
+              umount "$mountpoint"
+            done
+
+            esp_uuid="$(sed 's/\x00//' /proc/device-tree/chosen/asahi,efi-system-partition)"
+            # nixos-rebuild switch means that the mountpoint might be already mounted
```
flokli, line 76: If we do this at initrd time we shouldn't have to be worried something else has the ESP mounted. If we keep wanting to do this past initrd, we probably should use `mcopy` (from `mtools`) to read the cpio file from the ESP, to avoid having to deal with concurrent mounts.

CodeWitchBella: ah, nice! I didn't know about mcopy, that indeed simplifies the whole thing

CodeWitchBella:
> Hey, thanks. I didn't take too close of a look yet, but can you elaborate why you chose not to run this in the initrd? Any specific obstacles, or simply a 'this was easier to debug'? Would you be up to moving it to the initrd?

Mostly this was easier to debug. Also AFAIK, nixos has two initrd setups - one with systemd and one with bunch of scripts, so I didn't really want to touch that. The thinking was "hey, I can do this, how to do it in one evening before I loose motivation" :-)

> 
> I'm asking because https://asahilinux.org/docs/platform/open-os-interop/#linux-specific specifically describes this as happening in the initrd, and only there we can also make use of the cpio files provided by the bootloader and skip having to mount the ESP to get it from there in (most) cases.

My read of that one is that initrd is the suggested way (in the spirit of as early as possible), but the requirements is "drivers and modules, load before udev".

I might try moving to initrd, but I'm still hesitant given the double initrd nixos state. Unless we'll only support one variant?

jakobkukla:
Since systemd initrd is now the default in NixOS, how about only supporting this and checking with an assertion?

flokli:
> Since systemd initrd is now the default in NixOS, how about only supporting this and checking with an assertion?

SGTM

